PR #2, opened 2022-03-25
[WOR-91] Azure managed apps listing endpoint

aherbst-broad:
## Context
[Relevant ticket](https://broadworkbench.atlassian.net/browse/WOR-91)
We need to be able to return the set of azure managed apps a user has access to so they can choose one for creating a billing profile.

## This PR
* Creates a new endpoint at `/api/azure/v1/managedApps` that, when given a subscription ID, will reach out to Azure and pull in the managed applications a user has access to
* Introduces the notion of a "well known" azure marketplace offer to the BPM config. These offers will be loaded at runtime and will be the only ones against which a billing profile may be provisioned

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -22,7 +22,9 @@ dependencies {
 	implementation platform('com.google.cloud:libraries-bom:20.9.0')
 
 	// Cloud Resource Library
-	implementation  group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.1"
+	implementation group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.1"
+	implementation group: "com.azure.resourcemanager", name: "azure-resourcemanager-subscription", version: "1.0.0-beta.1"
+
 
 	// TODO: remove SNAPSHOT once https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/109 is merged
 	implementation platform('bio.terra.cloud-resource-lib:platform:0.8.2-SNAPSHOT')
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -22,6 +22,8 @@ dependencies {
 	implementation platform('com.google.cloud:libraries-bom:20.9.0')
 
 	// Cloud Resource Library
+	implementation  group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.1"
+
 	// TODO: remove SNAPSHOT once https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/109 is merged
 	implementation platform('bio.terra.cloud-resource-lib:platform:0.8.2-SNAPSHOT')
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'google-billing'
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -9,17 +9,9 @@ plugins {
 apply from: 'generators.gradle'
 apply from: 'publishing.gradle'
 
-repositories {
-	mavenCentral()
-	google()
-	maven {
-		url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
-	}
-}
-
 dependencies {
-	implementation  group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.1"
-	implementation 'bio.terra:terra-common-lib'
+	implementation 'bio.terra:terra-common-lib:0.0.53-SNAPSHOT'
+
 	implementation 'org.apache.commons:commons-dbcp2'
 	implementation 'org.springframework.boot:spring-boot-starter-data-jdbc'
 	implementation 'org.springframework.boot:spring-boot-starter-web'
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -9,7 +9,16 @@ plugins {
 apply from: 'generators.gradle'
 apply from: 'publishing.gradle'
 
+repositories {
+	mavenCentral()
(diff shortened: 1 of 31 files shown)


service/build.gradle
```diff
@@ -9,7 +9,16 @@ plugins {
 apply from: 'generators.gradle'
 apply from: 'publishing.gradle'
 
+repositories {
```
rtitle, line 12: Could we add the repo here instead if it's not currently included? https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/buildSrc/src/main/groovy/bio.terra.profile.java-common-conventions.gradle#L25

service/build.gradle
```diff
@@ -9,7 +9,16 @@ plugins {
...
+	mavenCentral()
+	google()
+	maven {
+		url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
+	}
+}
+
 dependencies {
+	implementation  group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.1"
```
rtitle, line 21: Normally we pull in cloud libraries through CRL. We still use the cloud libraries directly, but CRL is able to inject cleanup/logging for resource creations, etc.

BPM is not creating any cloud resources though, so I feel less strongly that it must go through CRL… so this seems ok I think.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -277,6 +302,22 @@ components:
...
 
+    AzureManagedAppsResponseModel:
+      type: object
+      properties:
+        managedApps:
+          type: array
+          items:
+            $ref: '#/components/schemas/AzureManagedAppModel'
+
+    AzureManagedAppModel:
```
rtitle, line 313: Wondering if (tenant, subscription, name) should be in the response, since presumably the caller will use this info to create a billing profile?

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
```diff
@@ -0,0 +1,58 @@
...
+   * @param userRequest Authorized user request
+   * @return List of Terra Azure managed applications the user has access to
+   */
+  public List<AzureManagedAppModel> getManagedAppDeployments(
+      UUID subscriptionId, AuthenticatedUserRequest userRequest) {
+    var appMgr = this.crlService.getApplicationManager(subscriptionId);
+    return appMgr.applications().list().stream()
+        .filter(app -> isAuthedTerraManagedApp(userRequest, app))
+        .map(app -> new AzureManagedAppModel().name(app.name()))
+        .toList();
```
rtitle, line 41: I read in JDK17 you didn't need `.collect(Collectors.toList())` to convert a Stream to a List but it didn't work in my IDE ;-)

rtitle:
Few comments/questions above, overall looks good though. 👍

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
pshapiro4broad: The form here (`.toList()`) is the new one. If it doesn't work in the IDE maybe you have a project setting that isn't enabling JDK 17 features?

service/src/test/java/bio/terra/profile/service/azure/AzureServiceUnitTest.java
```diff
@@ -0,0 +1,80 @@
...
+    var user =
+        AuthenticatedUserRequest.builder()
+            .setSubjectId("12345")
+            .setEmail(authedUserEmail)
+            .setToken("token")
+            .build();
+    var offerName = "known_terra_offer";
+
+    var authedTerraApp = mock(Application.class);
+    when(authedTerraApp.plan()).thenReturn(new Plan().withProduct(offerName));
```
aherbst-broad, line 35: I am not thrilled with the amount of mocking I'm doing here. The alternative is to introduce an adapter class that consumes a com.azure.Application and produces a more testable type. I think this should be ok for now though.

service/src/test/java/bio/terra/profile/service/azure/AzureServiceUnitTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    var result = azureService.getManagedAppDeployments(subId, user);
+
+    var expected =
+        List.of(
+            new AzureManagedAppModel()
+                .tenantId(tenantId)
+                .managedResourceGroupId("mrg_fake1")
+                .subscriptionId(subId));
+    assertEquals(result, expected);
```
pshapiro4broad, line 78: This assert is backwards; in JUnit assert the `expected` value comes first.
```suggestion
    assertEquals(expected , result);
```

cahrens (approved):
I've read through all the code and it mostly makes sense to me. I just find it a bit difficult to keep in my head how all the different service/controller classes relate to each other. If you have a diagram, please point me to it.

marctalbott (approved):

service/src/main/java/bio/terra/profile/app/StartupInitializer.java
```diff
@@ -14,6 +15,8 @@ public static void initialize(ApplicationContext applicationContext) {
     JobService jobService = applicationContext.getBean(JobService.class);
     ProfileDatabaseConfiguration profileDatabaseConfiguration =
         applicationContext.getBean(ProfileDatabaseConfiguration.class);
+    AzureConfiguration azureConfiguration = applicationContext.getBean(AzureConfiguration.class);
+    azureConfiguration.logOffers();
```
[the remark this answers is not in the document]
aherbst-broad: This is not for testing, I find it helpful to have these log out at startup. We can remove downstream if it's too verbose.

PR #3, opened 2022-03-31
Plug spring vulnerability

rtitle:
See see https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement

Also added dependency locking, since that wasn't present in this repo previously.


rtitle:
Build is complaining about missing repository, will look more later… this isn't urgent because this service is not in prod yet.

PR #4, opened 2022-05-10
[WOR-315] Add publish action

aherbst-broad:
## Context
Getting CI setup for BPM and ready for a dev deployment

## This PR
* Adds a publish action that fires on a new tag
* Adds a BPM specific slack hook instead of reusing the data catalog one

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,6 +25,9 @@ jobs:
             ~/.gradle/caches
             ~/.gradle/wrapper
           key: v1-${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            v1-${{ runner.os }}-gradle-
+
       - name: Parse tag
         id: tag
         run: echo ::set-output name=tag::$(git describe --tags)
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,7 +25,6 @@ jobs:
             ~/.gradle/caches
             ~/.gradle/wrapper
           key: v1-${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
-
       - name: Parse tag
         id: tag
         run: echo ::set-output name=tag::$(git describe --tags)
@@ -61,14 +60,14 @@ jobs:
         run: docker push ${{ steps.image-name.outputs.name }}
 
     # TODO dispatch notification to terra-helmfile once a chart is setup
-
-      - name: Notify slack on failure
-        uses: broadinstitute/action-slack@v3.8.0
-        if: failure()
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.BPM_SLACK_WEBHOOK }}
-        with:
-          status: ${{ job.status }}
-          author_name: Publish to dev
-          fields: job
-          text: 'Publish failed :sadpanda:'
+#
+#      - name: Notify slack on failure
+#        uses: broadinstitute/action-slack@v3.8.0
+#        if: failure()
+#        env:
+#          SLACK_WEBHOOK_URL: ${{ secrets.BPM_SLACK_WEBHOOK }}
+#        with:
+#          status: ${{ job.status }}
+#          author_name: Publish to dev
+#          fields: job
+#          text: 'Publish failed :sadpanda:'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,6 +7,7 @@ env:
 
 jobs:
   publish-job:
+    # only publish on new tag
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-refs/heads/dev-${{ github.sha }}
```
pshapiro4broad, line 27: I'm wondering about what this job is doing, since we don't use it in the catalog workflow code. Does it speed up the build? I remember looking at gradle build caching when I was setting up catalog and ended up looking at https://github.com/burrunan/gradle-cache-action but haven't done anything with it yet.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+        run: echo ::set-output name=name::gcr.io/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ steps.tag.outputs.tag }}
+
+      - name: Add Google Cloud Profiler to Docker Image
+        run: docker build ./service -t billingprofilemanager:local
+
+      - name: Build image locally with jib
+        run: |
+          ./gradlew --build-cache :service:jibDockerBuild \
+          --image=${{ steps.image-name.outputs.name }} \
+          -Djib.from.image=docker://billingprofilemanager:local \
```
pshapiro4broad, line 58: I'm wondering about this too, since we don't use this directive when calling `jib` in our workflow. I thought that `jib.from.image` was a way to specify the base image. Does this override https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/publishing.gradle#L24 ?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,75 @@
...
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            v1-${{ runner.os }}-gradle-
+
+      - name: Parse tag
+        id: tag
+        run: echo ::set-output name=tag::$(git describe --tags)
+
+      # TODO publish client
```
aherbst-broad, line 35: Ticket this

.github/workflows/publish.yml
```diff
@@ -0,0 +1,75 @@
...
+        with:
+          credentials_json: ${{ secrets.GCR_PUBLISH_KEY_B64 }}
+      - name: Explicitly auth Docker for GCR
+        run: gcloud auth configure-docker --quiet
+
+      - name: Construct docker image name and tag
+        id: image-name
+        run: echo ::set-output name=name::gcr.io/${GOOGLE_PROJECT}/${SERVICE_NAME}:${{ steps.tag.outputs.tag }}
+
+      # TODO add google cloud profiler
```
aherbst-broad, line 48: Ticket this

.github/workflows/publish.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Run Trivy vulnerability scanner
+        uses: broadinstitute/dsp-appsec-trivy-action@v1
+        with:
+          image: ${{ steps.image-name.outputs.name }}
+
+      - name: Push GCR image
+        run: docker push ${{ steps.image-name.outputs.name }}
+
+      # TODO dispatch notification to terra-helmfile once a chart is setup
```
aherbst-broad, line 64: Ticket this

rtitle (approved):
This LGTM, I see a successful publish action 👍

.github/workflows/publish.yml
```diff
@@ -0,0 +1,75 @@
...
+on: create
+
+env:
+  SERVICE_NAME: ${{ github.event.repository.name }}
+  GOOGLE_PROJECT: broad-dsp-gcr-public
+
+jobs:
+  publish-job:
+    # only publish on new tag
+    if: startsWith(github.ref, 'refs/tags/')
```
aherbst-broad, line 11: Should only fire on main

.github/workflows/publish.yml
pshapiro4broad: I wonder how much improvement you'll get. I guess that would depend on how much time it takes to build in the first place.

.github/workflows/publish.yml
pshapiro4broad: I think BPM uses the same setup that catalog does, where the gradle config for `jib` handles downloading the cloud profiler and adding it to the base image. The BPM code for this is here: https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/publishing.gradle#L10

Given that setup, I think `jibDockerBuild` would be sufficient by itself.

pshapiro4broad: But maybe this form is better than the `jib` method? Or maybe it's doing something different altogether.

.github/workflows/publish.yml
aherbst-broad: https://broadworkbench.atlassian.net/browse/WOR-320

.github/workflows/publish.yml
aherbst-broad: https://broadworkbench.atlassian.net/browse/WOR-321

PR #7, opened 2022-05-24
[WOR-321] Update BPM helmfile after publishing

marctalbott:
Ticket: [WOR-321](https://broadworkbench.atlassian.net/browse/WOR-321)
* This is most similar to the [Data Catalog `publish.yml` action](https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/publish.yml) and will deploy new tags to the dev environment. It seems to me that this job should only run on pushes to the main branch similar to how [WSM does it](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/.github/workflows/tag-publish.yml), but that's a larger change so starting with this for now

PR #8, opened 2022-05-25
[WOR-321] trigger publish off successful test runs not new tags

marctalbott:
ticket: [WOR-321](https://broadworkbench.atlassian.net/browse/WOR-321)
* updates how the publish action is triggered. Note that it can still be run manually, but will no longer run automatically when new tags are created

1. `build-and-test` action is run on merges to `main` branch and when new PRs are opened
2. If `build-and-test` action succeeds AND it was run on a merge to the `main` branch, then the `Tag, deploy, and publish` action (publish.yml) will be dispatched
3. This action first tags the current HEAD of the `main` branch using the tag.yml action
4. Then it goes through the rest of the already established steps to build and push the image as well as deploy the image to the Terra dev environment using the changes found in https://github.com/DataBiosphere/terra-billing-profile-manager/pull/7

* There might still be some work to do to determine if we _need_ a new tag. If a new tag is needed, we will also probably want to provide the option to specify what level of version bump is required.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,7 +1,7 @@
 name: Tag
 on:
-  - workflow_dispatch
-  - workflow_call:
+  workflow_dispatch
+  workflow_call:
       outputs:
         new_tag:
           description: "Generated tag"
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,6 +1,6 @@
 name: Tag
 on:
-  workflow_dispatch
+  workflow_dispatch:
   workflow_call:
       outputs:
         new_tag:
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,9 +1,17 @@
 name: Tag
-on: workflow_dispatch
+on:
+  - workflow_dispatch
+  - workflow_call:
+      outputs:
+        new_tag:
+          description: "Generated tag"
+          value: ${{ jobs.tag-job.outputs.new_tag }}
 
 jobs:
   tag-job:
     runs-on: ubuntu-latest
+    outputs:
+      new_tag: ${{ steps.tag.outputs.new_tag }}
     steps:
       - name: Checkout current code
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/tag.yml
```diff
@@ -16,7 +24,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
           HOTFIX_BRANCHES: hotfix.*
-          DEFAULT_BUMP: minor
+          DEFAULT_BUMP: patch
```
cahrens, line 27: I'm not clear how this works-- can you explain a bit more? What will the automatic tag numbers be? If we need to change it to be something/specific (major release for instance), how do we do that?

cahrens: Would one manually create a tag and then manually run a build? I know how to create a tag in github, but I'm not clear how to trigger the build step.

aherbst-broad (approved):
lgtm

.github/workflows/tag.yml
pshapiro4broad: You can manually trigger any GitHub workflow that has a workflow event dispatch trigger, using the Run workflow dropdown on the workflow page: https://github.com/DataBiosphere/terra-billing-profile-manager/actions/workflows/tag.yml

In the case of this workflow (tag), you typically wouldn't run it manually. It's run a from the `build and test` workflow if that workflow was run on the `main` branch, which normally happens after a PR is merged.

Regarding `minor` vs `patch`; we started out using `patch` but then changed it as we reasoned that `minor` would be used for regular (non-major) releases, and `patch` would be used for patches only, where the only change was to fix an issue in a previous version.

To control the version number used for the release, you can use a commit message to override the `DEFAULT_BUMP` behavior. See https://github.com/DataBiosphere/github-actions/blob/master/actions/bumper/README.md#bumping

cahrens: Thanks @pshapiro4broad . @marctalbott do you think we should stick with `minor` here?

marctalbott: `patch` felt more natural to me as it's what we use in WSM and workbench-libs, so that's why I changed it. I don't feel particularly strongly about it though, so we can revert to `minor` if that's preferred

pshapiro4broad: (Regarding `minor` vs `patch`, both were discussed on our team so I wasn't trying to say what you have is incorrect, I wanted to let you know the thinking behind what we did.)

cahrens (approved):

.github/workflows/tag.yml
cahrens: I don't really know which one we want, but it's easy enough to change later if we wish.

PR #12, opened 2022-07-13
[WOR-389] Add last_modified to Billing Profile Manager (BPM)

yunip20:
Adds last_modified to billing profile manager

https://broadworkbench.atlassian.net/browse/WOR-389

### DIFF
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -318,6 +318,9 @@ components:
         createdDate:
           type: string
           description: Date the profile was created
+        lastModified:
+          type: string
+          description: Date the profile was last modified
         createdBy:
           type: string
           description: Identity of the user who originally created the profile
--- a/service/src/main/resources/api/service_openapi.yaml
+++ b/service/src/main/resources/api/service_openapi.yaml
@@ -320,7 +320,7 @@ components:
           description: Date the profile was created
         lastModified:
           type: string
-          description: Date the profile was last modified
+          description: Timestamp in UTC the profile was last modified
         createdBy:
           type: string
           description: Identity of the user who originally created the profile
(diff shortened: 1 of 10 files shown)


aherbst-broad:
Thanks for doing this, I'll take a look! Somewhat related: I typically include the ticket in the branch name for easy reference (i.e., if I'm working on it, put it aside and come.back to it a day later). This makes picking out the correct branch to switch back to much easier (something like yp-wor-389-blah-blah). No need to change the name of this branch though.

yunip20:
I understand, thank you!

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -318,6 +318,9 @@ components:
         createdDate:
           type: string
           description: Date the profile was created
+        lastModified:
+          type: string
+          description: Date the profile was last modified
```
aherbst-broad, line 323: The documentation here isn't quite correct, this is the timestamp in UTC the profile was last modifed, not the date.

service/src/main/resources/db/changesets/20220713_billing_profiles_last_modified.yaml
```diff
@@ -0,0 +1,14 @@
...
+      changes:
+        - addColumn:
+            tableName: billing_profile
+            columns:
+              - column:
+                  name: last_modified
+                  type: timestamptz
+                  defaultValueComputed: now()
+                  remarks: |
+                    Date the profile was last modified.
```
aherbst-broad, line 14: Nitpick: add a line break here

aherbst-broad (commented):
Looking good, had a few suggestions. I think there are also some linting issues judging by the CI failure.

service/src/main/resources/api/service_openapi.yaml
aherbst-broad: @yunip20 can you update this documentation as well?

aherbst-broad (approved):

PR #13, opened 2022-07-20
[WOR-390] Create Profile Sync

yunip20:
Updates create profile to be synchronous 

https://broadworkbench.atlassian.net/browse/WOR-390


service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
```diff
@@ -74,18 +74,18 @@ public void before() {
   @Test
```
yunip20, line 74: Looking for feedback on test coverage for this change. Do you have any suggestions?

cahrens (approved):

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
aherbst-broad: This seems like a reasonable to the test given your changes above.

aherbst-broad (approved):

PR #15, opened 2022-08-03
[WOR-431] Adjust azure profile data model managed app coordinates

aherbst-broad:
## Context
There are a couple of problems with the current create profile code path with respect to Azure. 
It pulls the app down for access control checks by using the application deployment name and the parent resource group name. This is problematic because we:
1. Should be using the same code path for access control checks when enumerating managed apps via the /managedApps endpoint as we are when creating a profile
2. We need the managed resource group ID to be persisted with the profile so we can provide that to WSM downstream when creating workspace resources.

## This PR
* Refactors the data model to swap out `applicationDeploymentName` + `resourceGroupName` for `managedResourceGroupId`
* Changes the profile creation logic to reuse the managed application enumeration codepath in `AzureService` as the `/managedApps` endpoint uses
* Introduces a `BaseUnitTest` base class for use when testing POJOs. In my testing, it was becoming quite cumbersome/slow to spin up spring boot tests for basic business logic tests.

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/azure/AzureService.java
+++ b/service/src/main/java/bio/terra/profile/service/azure/AzureService.java
@@ -33,7 +33,7 @@ public class AzureService {
    * @param userRequest Authorized user request
    * @return List of Terra Azure managed applications the user has access to
    */
-  public List<AzureManagedAppModel> getManagedAppDeployments(
+  public List<AzureManagedAppModel> getAuthorizedManagedAppDeployments(
       UUID subscriptionId, AuthenticatedUserRequest userRequest) {
     var tenantId = appService.getTenantForSubscription(subscriptionId);
 
@@ -49,6 +49,7 @@ public class AzureService {
                     .managedResourceGroupId(
                         normalizeManagedResourceGroupId(app.managedResourceGroupId()))
                     .tenantId(tenantId))
+        .distinct()
         .toList();
   }
 
--- a/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
+++ b/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
@@ -21,7 +21,7 @@ record CreateProfileVerifyDeployedApplicationStep(
     try {
       var apps =
           azureService.getAuthorizedManagedAppDeployments(profile.subscriptionId().get(), user);
-      var matchingApps =
+      canAccess =
           apps.stream()
               .filter(
                   app ->
@@ -29,8 +29,7 @@ record CreateProfileVerifyDeployedApplicationStep(
                               app.getManagedResourceGroupId(),
                               profile.managedResourceGroupId().get())
                           && app.getSubscriptionId() == profile.subscriptionId().get())
-              .toList();
-      canAccess = matchingApps.size() == 1;
+                  .count() == 1;
     } catch (Exception e) {
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
     }
--- a/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
+++ b/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
@@ -1,44 +1,36 @@
 package bio.terra.profile.service.profile.flight.create;
 
 import bio.terra.common.iam.AuthenticatedUserRequest;
-import bio.terra.profile.service.crl.CrlService;
+import bio.terra.profile.service.azure.AzureService;
 import bio.terra.profile.service.profile.exception.InaccessibleApplicationDeploymentException;
 import bio.terra.profile.service.profile.model.BillingProfile;
 import bio.terra.stairway.FlightContext;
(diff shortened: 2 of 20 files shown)


marctalbott (approved):

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
```diff
@@ -49,6 +49,7 @@ public List<AzureManagedAppModel> getManagedAppDeployments(
                     .managedResourceGroupId(
                         normalizeManagedResourceGroupId(app.managedResourceGroupId()))
                     .tenantId(tenantId))
+        .distinct()
```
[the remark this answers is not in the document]
rtitle: Weird it happens it in the Java SDK as well, I was assuming it was a CLI bug

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
```diff
@@ -1,44 +1,36 @@
...
-          properties
-              .get("parameters")
-              .get("authorizedTerraUser")
-              .get("value")
-              .equalsIgnoreCase(user.getEmail());
+      var apps =
+          azureService.getAuthorizedManagedAppDeployments(profile.subscriptionId().get(), user);
+      var matchingApps =
+          apps.stream()
+              .filter(
```
rtitle, line 26: Nit - maybe `Stream.anyMatch` would work here

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
```diff
@@ -1,44 +1,36 @@
...
+                  app ->
+                      Objects.equals(
+                              app.getManagedResourceGroupId(),
+                              profile.managedResourceGroupId().get())
+                          && app.getSubscriptionId() == profile.subscriptionId().get())
+              .toList();
+      canAccess = matchingApps.size() == 1;
```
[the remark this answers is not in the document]
rtitle: Could maybe do `.filter(…).count() == 1` as well.

There is `anyMatch` too but that short-circuits after the first match so won't tell you there is exactly 1 match.

(Though I don't think you can have 2 resource groups with the same name in the same subscription anyway..)

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -250,12 +250,9 @@ components:
...
           description: Unique identifier of the Azure subscription
           format: uuid
-        resourceGroupName:
-          type: string
-          description: Name of the Azure resource group
-        applicationDeploymentName:
+        managedResourceGroupId:
           type: string
-          description: Name of the Azure application deployment
+          description: ID of the Azure managed resource group
```
rtitle, line 255: Does this expect an ID like `/subscriptions/3efc5bdf-be0e-44e7-b1d7-c08931e3c16c/resourceGroups/mrg-terra-integration-test-20211118`? Maybe an example in the YAML description would help.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -353,12 +350,9 @@ components:
...
           description: Unique identifier of the Azure subscription
           format: uuid
-        resourceGroupName:
-          type: string
-          description: Name of the Azure resource group
-        applicationDeploymentName:
+        managedResourceGroupId:
           type: string
-          description: Name of the Azure application deployment
+          description: Managed resource group ID of the Terra deployment
```
rtitle, line 355: Nit - maybe "Terra managed application deployment" (I read it Terra control plane deployment in my head)

service/src/main/resources/db/changesets/20220803_billing_profiles_managed_resource_group_id.yaml
```diff
@@ -0,0 +1,23 @@
...
+            tableName: billing_profile
+            columns:
+              - column:
+                  name: managed_resource_group_id
+                  type: text
+                  constraints:
+                    nullable: true
+                  remarks: |
+                    Azure Managed Resource Group ID for the Terra deployment
+        - dropColumn:
```
rtitle, line 16: Are there existing records in the DB to worry about, which will not have a `managed_resource_group_id` populated? Or is it empty/do we not care.

rtitle (approved):
Nice! Some nits but LGTM.

Makes sense to standardize all services to use `(tenantId, subscriptionId, managedResourceGroupId)` as an Azure identifier.  I almost wonder if we could have some common Azure models in TCL or something.

service/src/main/resources/api/service_openapi.yaml
aherbst-broad: No, this is the bare ID i.e,. mrg-terra-integration-test-20211118. I'll update the documentation.

service/src/main/resources/db/changesets/20220803_billing_profiles_managed_resource_group_id.yaml
aherbst-broad: It is empty, we do not care.

aherbst-broad:
@rtitle yeah WSM is ambiguous in it's naming here as well (it refers to it as "resource group ID" but it is defacto the MRG ID). Agreed this would be good to get into TCL.

PR #24, opened 2022-08-16
WOR-414: Render BPM app azure client id, secret and tenant ID into local dev configs

blakery:
[WOR-414](https://broadworkbench.atlassian.net/browse/WOR-414)

update config script to pull service azure creds from vault to local config file

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,14 @@ After the database is initialized, run unit tests:
 ./gradlew test
 ```
 
+To set up serrvice account credentials and other configuration for running locally:
+* Install yq `brew install yq`
+* Run `render-configs.sh`
+
 To run integration tests:
 ```sh
 ./gradlew bootRun &    # start up a local instance of the billing profile manager service
 sleep 5                # wait until service comes up
-render-configs.sh      # render service account credentials needed for tests
 ./gradlew :integration:runTest --args="suites/FullIntegration.json /tmp/test"
 ```
 
--- a/render_configs.sh
+++ b/render_configs.sh
@@ -1,19 +1,21 @@
+#!/bin/bash
+
 ENV=${1:-dev}
 VAULT_TOKEN=${2:-$(cat "$HOME"/.vault-token)}
 
 VAULT_ADDR="https://clotho.broadinstitute.org:8200"
-COMMON_VAULT_PATH="secret/dsde/terra/kernel/$ENV/common"
 
-VAULT_COMMAND="vault read"
+VAULT_DOCKER_IMAGE="vault:1.11.2"
 
-INTEGRATION_OUTPUT_LOCATION="$(dirname "$0")/integration/src/main/resources/rendered"
+LOCAL_CONFIG_OUTPUT="service/src/main/resources/generated/local-properties.yaml"
 
-if ! [ -x "$(command -v vault)" ]; then
-  VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR vault:1.7.3 $VAULT_COMMAND"
-fi
+VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR $VAULT_DOCKER_IMAGE vault read"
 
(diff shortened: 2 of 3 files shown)


render_configs.sh
```diff
@@ -1,19 +1,21 @@
+#!/bin/bash
```
blakery, line 1: I changed this file as much as I did because the previous version didn't appear to be used anywhere:

- When run locally, the original script always fails, because it's trying to write configs to a location that doesn't exist in this project
- When deployed in real environments, the configs are passed via deployment infrastructure, not a locally rendered file

README.md
```diff
@@ -33,11 +33,14 @@ After the database is initialized, run unit tests:
 ./gradlew test
 ```
 
+To set up serrvice account credentials and other configuration for running locally:
+* Install yq `brew install yq`
```
blakery, line 37: `yq` provides basically the same functionality as `jq`, but for yaml instead of json. 

Since the vault file names don't line up with the property names expected by the service, we need to do some manipulation of the yaml. `yq` seemed the most lightweight way of doing that.

service/src/main/resources/application.yml
```diff
@@ -31,7 +31,7 @@ env:
 # contains the configuration of the BPM test application. We can use that application in our
 # integration testing to make sure the application code paths are working. However, we do not
 # want it to appear in production environments.
-spring.config.import: optional:file:../config/local-properties.yml
+spring.config.import: optional:file:./generated/local-properties.yml
```
blakery, line 34: Although this file location differs from that specified in workspace manager, there are a few reasons I think it makes sense to locate it here:

1. In practice, workspace manager doesn't appear to actually render anything into that location anyway
2. the `../config` location would put it in a folder located in `service/src/main`. I think this is better thought of as a resource
3. There was already a folder set up to be ignored by git at `service/src/main/resources/generated
4. This _is_ a generated resource

render_configs.sh
```diff
@@ -1,19 +1,21 @@
...
-  VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR vault:1.7.3 $VAULT_COMMAND"
-fi
+VAULT_COMMAND="docker run --rm -e VAULT_TOKEN=$VAULT_TOKEN -e VAULT_ADDR=$VAULT_ADDR $VAULT_DOCKER_IMAGE vault read"
 
-$VAULT_COMMAND -field=data -format=json "secret/dsde/firecloud/$ENV/common/firecloud-account.json" >"$INTEGRATION_OUTPUT_LOCATION/user-delegated-sa.json"
+function azure_creds {
+  AZURE_VAULT_PATH="secret/dsde/terra/azure/dev/billingprofilemanager/managed-app-publisher"
+  $VAULT_COMMAND -format=yaml $AZURE_VAULT_PATH \
+  | yq '.out.env.azure.managedAppClientId = .data.client-id | .out.env.azure.managedAppClientSecret = .data.client-secret  | .out.env.azure.managedAppTenantId = .data.tenant-id | .out' \
+  > $LOCAL_CONFIG_OUTPUT
```
blakery, line 18: If we add other resources in the future, we could just append to a local variable, and write it all out once at the end

render_configs.sh
marctalbott: The script didn't work for me as is. I was getting the following error message 
```
Couldn't start vault with IPC_LOCK. Disabling IPC_LOCK, please use --cap-add IPC_LOCK
```
I was able to fix it by adding `--cap-add=IPC_LOCK` flag to the `docker run` command.

An alternative would be to use the `dsde-toolbox` to access vault instead. You could replace `VAULT_COMMAND` with
```
docker run --rm -i -e VAULT_TOKEN="${VAULT_TOKEN}" broadinstitute/dsde-toolbox:dev vault read
```

blakery: It showed the same message for me, but still ran successfully. I assumed it was related to having low available resources available on my machine.

Switching to the second suggestion since it seems a bit cleaner.

marctalbott: 🤦  I was so concerned by the error message I didn't even think to check if it worked (it did). Oh well... thanks for making the change anyways

marctalbott (approved):

yunip20 (approved):

PR #25, opened 2022-08-16
Reworking well-known offers

aherbst-broad:
## Context
[Relevant ticket](https://broadworkbench.atlassian.net/jira/software/c/projects/WOR/boards/125/backlog?view=detail&selectedIssue=WOR-454&issueLimit=100)

The way "well-known" Azure marketplace offers is configured in BPM is awkward and difficult to modify via helmfiles. 

(Quick refresher: well known offers are those Azure application offers whose deployments BPM will allow to be linked to billing profiles)

Specifically, having the offer name be a key and the auth'ed user key be a value makes setting the config via OS environment variables intractable in my research.

## This PR
* Restructures how we configure well known application offers. The config is now an array of offer objects, with the keys being the offer name and the authed user key. The knock-on effect here is that we are able to set the offers via environment variables like so:
   * `PROFILE_AZURE_APPLICATIONOFFERS_0_NAME=foo;PROFILE_AZURE_APPLICATIONOFFERS_0_AUTHORIZEDUSERKEY=bar`
* [or via a helmfile change](https://github.com/broadinstitute/terra-helmfile/pull/3223/files)

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
@@ -2,7 +2,7 @@ package bio.terra.profile.app.configuration;
 
 import com.azure.core.credential.TokenCredential;
 import com.azure.identity.ClientSecretCredentialBuilder;
-import java.util.Map;
+import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -12,7 +12,7 @@ public record AzureConfiguration(
     String managedAppClientId,
     String managedAppClientSecret,
     String managedAppTenantId,
-    Map<String, AzureApplicationOffer> applicationOffers) {
+    Set<AzureApplicationOffer> applicationOffers) {
   private static final Logger logger = LoggerFactory.getLogger(AzureConfiguration.class);
 
   /**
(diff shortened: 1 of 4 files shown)


service/src/test/java/bio/terra/profile/service/azure/AzureServiceUnitTest.java
```diff
@@ -106,9 +107,9 @@ public void getManagedApps_dedupesApps() {
     when(appService.getTenantForSubscription(subId)).thenReturn(tenantId);
 
     var offer = new AzureConfiguration.AzureApplicationOffer();
-    offer.setName("FAKE_APP_NAME");
```
aherbst-broad, line 109: This test was using a different value for the offer key than the offer name, which was technically possible given the earlier map implementation.

service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
```diff
@@ -12,7 +12,7 @@ public record AzureConfiguration(
     String managedAppClientId,
     String managedAppClientSecret,
     String managedAppTenantId,
-    Map<String, AzureApplicationOffer> applicationOffers) {
+    Set<AzureApplicationOffer> applicationOffers) {
```
aherbst-broad, line 15: Offers are now exposed as a `Set`, which has the desirable effect of preventing a disconnect between the offer key and the offer name (see change in the test further on in this PR).

service/src/main/resources/application.yml
```diff
@@ -89,7 +89,7 @@ profile:
     managed-app-tenant-id: ${env.azure.managedAppTenantId}
 
     application-offers:
-      fake-terra-app:
+      - name: terra-workspace-dev
```
aherbst-broad, line 92: I think `terra-workspace-dev` should become the default offer BPM allows in DEV, see this [thread](https://broadinstitute.slack.com/archives/C02GJ782BAA/p1660665478963609)

cahrens (approved):
Thanks for the cleanup-work, @aherbst-broad!

blakery (approved):
More clear and easier to work with.

PR #28, opened 2022-08-18
[WOR-456] run sonarqube scan in build-and-test

marctalbott:
Ticket: [WOR-456](https://broadworkbench.atlassian.net/browse/WOR-456)
* still need to add the `SONAR_TOKEN` secret to the repo
* follows [Appsec documentation](https://app.gitbook.com/o/-LVj1clrBdFW94pG2z2w/s/-LsrggqZs3CYGFRhFZdX-887967055/appsec-team-internal/appsec-team-internal/security-activities/sast-1) and configuration found in ECM and DrsHub

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -27,21 +27,8 @@ jobs:
           key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
           restore-keys: v1-${{ runner.os }}-gradle-${{ github.ref }}
 
-      - name: Cache SonarCloud packages
-        uses: actions/cache@v2
-        with:
-          path: ~/.sonar/cache
-          key: ${{ runner.os }}-sonar
-          restore-keys: ${{ runner.os }}-sonar
-
       - name: Gradle build service
         run: ./gradlew --build-cache :service:build -x test
-
-      - name: SonarQube scan
-        run: ./gradlew --build-cache :service:sonarqube
-        env:
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload spotbugs results
         uses: github/codeql-action/upload-sarif@main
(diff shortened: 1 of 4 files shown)


TomConner:
Secret is coming soon, just needs some Atlantis

TomConner:
Optional: add a badge to README.md…  `[Quality Gate Status](https://sonarcloud.io/summary/new_code?id=DataBiosphere_terra-billing-profile-manager)`

TomConner:
Standby for https://github.com/broadinstitute/terraform-ap-deployments/pull/759

.github/workflows/build-and-test.yml
```diff
@@ -27,9 +27,22 @@ jobs:
...
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+
       - name: Gradle build service
         run: ./gradlew --build-cache :service:build -x test
 
+      - name: SonarQube scan
+        run: ./gradlew --build-cache :service:sonarqube
```
pshapiro4broad, line 41: Running sonarqube this way won't pick up coverage information from your build. If you want coverage data as part of the sonar scan, you must run it after running `test`.

The good news is, once you've done this you can remove `codecov` from your workflow.

For an example, see https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/build-and-test.yml#L121

.github/workflows/build-and-test.yml
```diff
@@ -27,9 +27,22 @@ jobs:
           key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
           restore-keys: v1-${{ runner.os }}-gradle-${{ github.ref }}
 
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v2
```
pshapiro4broad, line 31: Is there a performance benefit for using this cache action? We don't use `actions/cache` in our workflows and I'm wondering if we should start doing it. The sonar scan in our build takes less than 1min so it's already pretty fast.

.github/workflows/build-and-test.yml
pshapiro4broad: It looks like you also need to configure XML coverage output from `jacoco` if you aren't doing that already.

https://github.com/DataBiosphere/terra-data-catalog/blob/main/buildSrc/src/main/groovy/bio.terra.catalog.java-common-conventions.gradle#L97

.github/workflows/build-and-test.yml
TomConner: Whatever works. This came in with the template action provided by SonarCloud but you're welcome to change it. I don't see why there would be a benefit to caching actually if it's running one scan on an action runner that's going to be thrown away.

.github/workflows/build-and-test.yml
TomConner: Good ideas! Perhaps I can update the recommendations. From AppSec's point of view, we don't require coverage reporting but it is nice in the sonarqube status, and if it is better than codecov then hey why not.

TomConner (approved):
This is working now, but at your discretion, you may want to remove caching or add coverage reporting as discussed.

aherbst-broad (approved):

PR #30, opened 2022-08-19
[WOR-487] setup minniekenny so building BPM will ensure that git-secrets is installed

marctalbott:
Ticket: [WOR-487](https://broadworkbench.atlassian.net/browse/WOR-487)
* read more about minnie-kenny here: https://minnie-kenny.readthedocs.io/en/latest/

### DIFF
--- a/minnie-kenny.gitconfig
+++ b/minnie-kenny.gitconfig
@@ -1,0 +1,11 @@
+[secrets]
+    providers = git secrets --aws-provider
+    patterns = (A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}
+    patterns = (\"|')?(AWS|aws|Aws)?_?(SECRET|secret|Secret)?_?(ACCESS|access|Access)?_?(KEY|key|Key)(\"|')?\\s*(:|=>|=)\\s*(\"|')?[A-Za-z0-9/\\+=]{40}(\"|')?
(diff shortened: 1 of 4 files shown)


minnie-kenny.gitconfig
```diff
@@ -0,0 +1,11 @@
+[secrets]
+    providers = git secrets --aws-provider
+    patterns = (A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}
+    patterns = (\"|')?(AWS|aws|Aws)?_?(SECRET|secret|Secret)?_?(ACCESS|access|Access)?_?(KEY|key|Key)(\"|')?\\s*(:|=>|=)\\s*(\"|')?[A-Za-z0-9/\\+=]{40}(\"|')?
+    patterns = (\"|')?(AWS|aws|Aws)?_?(ACCOUNT|account|Account)_?(ID|id|Id)?(\"|')?\\s*(:|=>|=)\\s*(\"|')?[0-9]{4}\\-?[0-9]{4}\\-?[0-9]{4}(\"|')?
+    allowed = AKIAIOSFODNN7EXAMPLE
```
aherbst-broad, line 6: Do we actually want these to be allowed?

aherbst-broad (approved):

blakery (approved):

minnie-kenny.gitconfig
marctalbott: I guess these are common example keys used in AWS that are added to the git-secrets config when you add the AWS provider which is built into git secrets. So these allowed strings and some of the other blocked AWS patterns are probably unnecessary because we don't do AWS things, but it's sort of the default git secrets config so I'm inclined to leave it.

At some point we could consider adding Azure-specific patterns, but since we only deal with GCP SAs at the moment I don't think it's high priority. There isn't a built in Azure provider in git secrets either which makes it a bit more difficult as I'm not sure what Azure secrets we would even want to catch

PR #38, opened 2022-09-14
[WOR-391] Implement Sam passthroughs for profile policies

cahrens:
Implements these endpoints for policies "user" and "owner", which map to roles "user" and "owner" respectively.

* GET /api/profiles/v1/{profileId}/policies

List the Sam resource policies for a billing profile.

* POST ​/api​/profiles/v1​/{profileId}​/policies​/{policyName}​/members​/{memberEmail}

Adds a member to the specified policy for the billing profile

* DELETE ​/api​/profiles/v1​/{profileId}​/policies​/{policyName}​/members​/{memberEmail}

Removes a member from the specified policy for the profile

The only information returned about a policy is its name and the member emails. For some discussion on this, see https://broadworkbench.atlassian.net/browse/WOR-391?focusedCommentId=66574

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -35,7 +35,7 @@ dependencies {
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'azure-resourcemanager-common'
 
 	// Sam
-	implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.12", version: "0.1-61135c7"
+	implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.13", version: "0.1-c53306a-SNAP"
 
 	// Test deps
 	testImplementation('org.springframework.boot:spring-boot-starter-test') {
--- a/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
@@ -80,7 +80,7 @@ public class ProfileApiController implements ProfileApi {
   public ResponseEntity<SamPolicyModelList> getProfilePolicies(@PathVariable("profileId") UUID id) {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     List<SamPolicyModel> policies = profileService.getProfilePolicies(id, user);
-    var response = new SamPolicyModelList().items(new ArrayList<>(policies));
+    var response = new SamPolicyModelList().items(policies);
 
     return new ResponseEntity<>(response, HttpStatus.OK);
   }
--- a/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
@@ -9,7 +9,6 @@ import bio.terra.profile.service.job.JobService;
 import bio.terra.profile.service.profile.ProfileService;
 import bio.terra.profile.service.profile.model.BillingProfile;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
@@ -86,27 +85,25 @@ public class ProfileApiController implements ProfileApi {
   }
 
   @Override
-  public ResponseEntity<PolicyResponse> addProfilePolicyMember(
+  public ResponseEntity<SamPolicyModel> addProfilePolicyMember(
       @PathVariable("profileId") UUID id,
       @PathVariable("policyName") String policyName,
       @PathVariable("memberEmail") String memberEmail) {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     SamPolicyModel policy =
         profileService.addProfilePolicyMember(id, policyName, memberEmail, user);
-    PolicyResponse response = new PolicyResponse().policies(Collections.singletonList(policy));
-    return new ResponseEntity<>(response, HttpStatus.OK);
+    return new ResponseEntity<>(policy, HttpStatus.OK);
   }
 
   @Override
-  public ResponseEntity<PolicyResponse> deleteProfilePolicyMember(
+  public ResponseEntity<SamPolicyModel> deleteProfilePolicyMember(
       @PathVariable("profileId") UUID id,
       @PathVariable("policyName") String policyName,
       @PathVariable("memberEmail") String memberEmail) {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     SamPolicyModel policy =
         profileService.deleteProfilePolicyMember(id, policyName, memberEmail, user);
-    PolicyResponse response = new PolicyResponse().policies(Collections.singletonList(policy));
-    return new ResponseEntity<>(response, HttpStatus.OK);
+    return new ResponseEntity<>(policy, HttpStatus.OK);
   }
 
   /**
--- a/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
@@ -87,24 +87,24 @@ public class ProfileApiController implements ProfileApi {
 
   @Override
   public ResponseEntity<PolicyResponse> addProfilePolicyMember(
-         @PathVariable("profileId") UUID id,
-         @PathVariable("policyName") String policyName,
-         @PathVariable("memberEmail") String memberEmail
- ) {
-   AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
-   SamPolicyModel policy = profileService.addProfilePolicyMember(id, policyName, memberEmail, user);
-   PolicyResponse response = new PolicyResponse().policies(Collections.singletonList(policy));
-   return new ResponseEntity<>(response, HttpStatus.OK);
- }
+      @PathVariable("profileId") UUID id,
+      @PathVariable("policyName") String policyName,
+      @PathVariable("memberEmail") String memberEmail) {
+    AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
+    SamPolicyModel policy =
+        profileService.addProfilePolicyMember(id, policyName, memberEmail, user);
+    PolicyResponse response = new PolicyResponse().policies(Collections.singletonList(policy));
+    return new ResponseEntity<>(response, HttpStatus.OK);
+  }
 
   @Override
   public ResponseEntity<PolicyResponse> deleteProfilePolicyMember(
-          @PathVariable("profileId") UUID id,
-          @PathVariable("policyName") String policyName,
-          @PathVariable("memberEmail") String memberEmail
-  ) {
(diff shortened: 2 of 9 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -145,6 +145,80 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/SamPolicyModelList'
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '403':
+          $ref: '#/components/responses/PermissionDenied'
+        '404':
+          $ref: '#/components/responses/NotFound'
+
+  /api/profiles/v1/{profileId}/policies/{policyName}/members/{memberEmail}:
```
cahrens, line 170: The jade-data-repo version of this supplies the `memberEmail` through the request policy instead of having it in the URL (for POST, not DELETE). I don't see a compelling reason to do that though (unless we plan to bulk add users), and prefer to keep this method symmetric with the DELETE endpoint.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -145,6 +145,80 @@ paths:
...
+        - $ref: '#/components/parameters/ProfileId'
+        - $ref: '#/components/parameters/ProfilePolicyName'
+        - $ref: '#/components/parameters/MemberEmail'
+      responses:
+        201:
+          description: Policy
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SamPolicyModel'
```
cahrens, line 188: The jade-data-repo versions return a list of policies, but given that there will only ever be one, that seems unnecessarily complex.

service/src/main/resources/application.yml
```diff
@@ -92,7 +92,7 @@ profile:
     managed-app-tenant-id: ${env.azure.managedAppTenantId}
 
     application-offers:
-      - name: terra-workspace-dev
+      - name: aherbst-202202024-preview
```
cahrens, line 95: @aherbst-broad is it OK to check this in?

service/build.gradle
```diff
@@ -35,7 +35,7 @@ dependencies {
 	implementation group: 'bio.terra.cloud-resource-lib', name: 'azure-resourcemanager-common'
 
 	// Sam
-	implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.12", version: "0.1-61135c7"
+	implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.13", version: "0.1-c53306a-SNAP"
```
cahrens, line 38: I updated the version of Sam Client because I initially was returning the policy entry memberPolicies, and it only exists in this newer version (I tried to update to the most recent few Sam commits, but that client JARs could not be found so it doesn't seem that a new JAR is generated on every commit). I am no longer returning anything that requires the updated Sam Client, but it seems good to still update the version and use more of the v2 endpoints.

service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
```diff
@@ -77,6 +75,37 @@ public ResponseEntity<Void> deleteProfile(@PathVariable("profileId") UUID id) {
     return new ResponseEntity<>(HttpStatus.NO_CONTENT);
   }
 
+  @Override
+  public ResponseEntity<SamPolicyModelList> getProfilePolicies(@PathVariable("profileId") UUID id) {
+    AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
```
cahrens, line 80: `AuthenticatedUserRequest` is deprecated, but we use it all through this codebase. Do we have a plan to change to `BearerToken`?

service/src/main/java/bio/terra/profile/service/iam/SamService.java
```diff
@@ -166,6 +163,84 @@ public List<UUID> listProfileIds(AuthenticatedUserRequest userRequest)
...
+      throw SamExceptionFactory.create("Error adding profile policy member in Sam", e);
+    }
+  }
+
+  private SamPolicyModel addProfilePolicyMemberInner(
+      AuthenticatedUserRequest userReq, UUID resourceId, String policyName, String userEmail)
+      throws ApiException {
+    ResourcesApi samResourceApi = samResourcesApi(userReq.getToken());
+    String samResourceName = SamResourceType.PROFILE.getSamResourceName();
+    logger.debug(
```
cahrens, line 203: Opinions wanted. Should I:

1. Make this "info" instead of "debug".
2. Not log it at all.
3. Leave as-is.

service/src/main/java/bio/terra/profile/service/iam/SamService.java
```diff
@@ -23,11 +24,7 @@
 import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.StatusApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi;
-import org.broadinstitute.dsde.workbench.client.sam.model.AccessPolicyMembershipV2;
-import org.broadinstitute.dsde.workbench.client.sam.model.CreateResourceRequestV2;
-import org.broadinstitute.dsde.workbench.client.sam.model.ResourceAndAccessPolicy;
-import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
-import org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo;
+import org.broadinstitute.dsde.workbench.client.sam.model.*;
```
marctalbott, line 27: I was asked to avoid `*` imports in WSM, do we want to do the same in BPM?

service/src/main/java/bio/terra/profile/service/iam/SamService.java
marctalbott: I'd be okay with not logging it at all, the Apache proxy logs will include this information since it's all part of the URL

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -145,6 +145,80 @@ paths:
...
+        '404':
+          $ref: '#/components/responses/NotFound'
+
+  /api/profiles/v1/{profileId}/policies/{policyName}/members/{memberEmail}:
+    post:
+      tags:
+        - Profile
+      description: >
+        Adds a member to the specified policy for the profile.
+        NOTE: Change may take up to 60 seconds to take effect.
```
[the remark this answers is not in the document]
marctalbott: The update to Sam should be instant, the only delay will be when synchronizing the policy membership with Google (if applicable). `spend-profile` resources will likely not be synced with Google regardless of the profile's cloud provider. It's fine to remove this comment imo

marctalbott (approved):
lgtm

service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
marctalbott: https://broadworkbench.atlassian.net/browse/WOR-510

service/src/main/java/bio/terra/profile/service/iam/SamService.java
cahrens: Hmmm, I personally think that `*` is fine, especially if there is nothing in place to prevent them from being used. If we are going to enforce this, it think it needs to me a spotless rule.

@aherbst-broad do you have an opinion?

service/src/main/resources/application.yml
cahrens: @aherbst-broad I'm assuming **no** given https://github.com/broadinstitute/terra-helmfile/pull/3312/files?

service/src/main/java/bio/terra/profile/service/iam/SamService.java
blakery: I also think `*` is fine. 
The original (as in, in the 1990s) reason to avoid it in Java was to avoid name collisions; this doesn't matter anymore, because Java supports the ability to use explicit imports to clarify those cases.
For example:
```Java
import java.util.*
import project.package.that.contains.something.called.list.*
import java.util.List
```
Static initialization happens when the class is loaded (before `main` is called), so unless you were dynamically loading custom jars or something weird, that wouldn't be a concern.

service/src/main/java/bio/terra/profile/service/iam/SamService.java
```diff
@@ -166,6 +163,84 @@ public List<UUID> listProfileIds(AuthenticatedUserRequest userRequest)
     }
   }
 
+  public List<SamPolicyModel> retrieveProfilePolicies(
```
blakery, line 166: Would it be worth it to make more liberal use of javadoc comments generally, but particularly for the primary entrypoints to a class? 

Not thinking anything fancy, but it can make things easier to work with from other classes, since the javadocs will show up in IDEs, as well as a few other knock-on benefits.

Not saying you should do anything about it here, but I figured it was a reasonable time to bring it up.

service/src/main/java/bio/terra/profile/service/iam/SamService.java
blakery: I agree that capturing it elsewhere seems adequate.

A lot of http clients have logging built in too, and ultimately, there's nothing going on here that you wouldn't be able to discern from the original request.

service/build.gradle
blakery: So long as our deployments aren't ahead of Sam, this seems fine to me.

blakery (approved):

service/src/main/java/bio/terra/profile/service/iam/SamService.java
cahrens: Will remove it.

service/src/main/java/bio/terra/profile/service/iam/SamService.java
cahrens: Good call, I will add JavaDoc to the public methods.

service/src/main/java/bio/terra/profile/service/iam/SamService.java
pshapiro4broad: > The original (as in, in the 1990s) reason to avoid it in Java was to avoid name collisions; this doesn't matter anymore, because Java supports the ability to use explicit imports to clarify those cases.

I've been using Java since the 1990s (I worked on one of the first Java IDEs) and I don't think I've ever heard this as a reason to avoid static imports. The reason I've heard (and repeated, because I believe it) is they should be avoided because it makes programs harder to read. This is why I would avoid both static imports and wildcard imports, with the exception being specific classes in test code (hamcrest and mockito APIs).

The [google java style guide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements), which DSP follows, says `No wildcard imports` and `No static import for classes` (although for static import it sounds like they're specifically talking about nested static classes).

An easy way to enforce this is to use an `.editorconfig` for IntelliJ that has rules to prevent wildcards from being used. Here's the file we use for catalog: https://github.com/DataBiosphere/terra-data-catalog/blob/main/.editorconfig

PR #39, opened 2022-09-16
[WOR-387] Switch email from URL param to body.

cahrens:
Switch the API to specify the email address in the  body because the Java client errors with no body present. I tried specifying an empty body, but that actually led to a less intuitive API, and I still couldn't get it to work (problems serializing an empty bean).

FWIW, this change makes the API match the [jade-data-repo one](https://github.com/DataBiosphere/jade-data-repo/blob/aa3320d38cf88eadfb31fca4ced3777f14080e29/src/main/resources/api/data-repository-openapi.yaml#L354).

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
@@ -88,10 +88,10 @@ public class ProfileApiController implements ProfileApi {
   public ResponseEntity<SamPolicyModel> addProfilePolicyMember(
       @PathVariable("profileId") UUID id,
       @PathVariable("policyName") String policyName,
-      @PathVariable("memberEmail") String memberEmail) {
+      @RequestBody PolicyMemberRequest requestBody) {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     SamPolicyModel policy =
-        profileService.addProfilePolicyMember(id, policyName, memberEmail, user);
+        profileService.addProfilePolicyMember(id, policyName, requestBody.getEmail(), user);
     return new ResponseEntity<>(policy, HttpStatus.OK);
   }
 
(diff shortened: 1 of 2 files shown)


service/src/main/resources/api/service_openapi.yaml
```diff
@@ -273,7 +278,7 @@ components:
     MemberEmail:
       name: memberEmail
       in: path
-      description: The email of the user to add or remove policies for
+      description: The email of the user to remove policies for
```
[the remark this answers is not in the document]
blakery: DELETE works the same as GET, in that proxies are not required to forward the body.

blakery (approved):

service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
```diff
@@ -88,10 +88,10 @@ public ResponseEntity<SamPolicyModelList> getProfilePolicies(@PathVariable("prof
   public ResponseEntity<SamPolicyModel> addProfilePolicyMember(
       @PathVariable("profileId") UUID id,
       @PathVariable("policyName") String policyName,
-      @PathVariable("memberEmail") String memberEmail) {
+      @RequestBody PolicyMemberRequest requestBody) {
```
pshapiro4broad, line 91: I'd recommend against copying down these REST annotations from the base class, as they have no benefit, only downsides. They _do_ override the values set in the base class, but aren't automatically updated when the base class version changes. So you end up with cases where a value changes in the yaml but it's not reflected in the code. The most common case is with a default value, where the value is changed in the base class interface but not changed in the subclass, so the openapi yaml change doesn't have an effect. That's why I think it's better to not annotate the subclasses at all.

cahrens: Thanks for the information @pshapiro4broad! I have removed them in https://github.com/DataBiosphere/terra-billing-profile-manager/pull/42.

PR #45, opened 2022-09-22
[WOR-424] Validate Azure resource providers

aherbst-broad:
## Context
We need to check that needed Azure [resource providers](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-providers-and-types) are registered in a subscription before creating a billing profile, otherwise Terra will be non-functional. 

## This PR
* Adds a check in the verification of the MRG step during profile creation that ensures all required providers are registered. If any are missing, we fail out with a 409.
   * We talked about using 424, but it looks like that is a WebDAV specific error code according to MDN.
* Adds the required providers to the application.yaml configuration.


service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
```diff
@@ -39,12 +46,27 @@ public StepResult doStep(FlightContext context) {
       throw new InaccessibleApplicationDeploymentException(
           String.format(
               "The user [%s] needs access to deployed application [%s] to perform the requested operation",
-              user.getEmail(), profile.managedResourceGroupId().get()));
+              user.getEmail(), profile.getRequiredManagedResourceGroupId()));
     }
 
+    checkRegisteredProviders();
```
aherbst-broad, line 52: Not doing this in the `try/catch` above as this is not a retriable check.

blakery (approved):

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
```diff
@@ -54,6 +61,26 @@ public List<AzureManagedAppModel> getAuthorizedManagedAppDeployments(
         .toList();
   }
 
+  /**
+   * Gets the resource providers in a subscription that are in either the "Registered" or
+   * "Registering" state.
```
cahrens, line 66: Out of curiosity, what other states could they be in?

service/src/main/java/bio/terra/profile/service/profile/exception/MissingRequiredProvidersException.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.profile.service.profile.exception;
+
+import bio.terra.common.exception.ErrorReportException;
+import java.util.List;
+import org.springframework.http.HttpStatus;
+
+public class MissingRequiredProvidersException extends ErrorReportException {
```
cahrens, line 7: Should you update the Swagger to document this possible exception from the create method?

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
```diff
@@ -54,6 +61,26 @@ public List<AzureManagedAppModel> getAuthorizedManagedAppDeployments(
...
+   */
+  public Set<String> getRegisteredProvidersForSubscription(UUID tenantId, UUID subscriptionId) {
+    var resourceManager = crlService.getResourceManager(tenantId, subscriptionId);
+    var providers = resourceManager.providers();
+    return providers.list().stream()
+        .filter(
+            provider ->
+                provider.registrationState().equals("Registered")
+                    || provider.registrationState().equals("Registering"))
+        .map(Provider::namespace)
```
cahrens, line 80: What does line 80 do? Convert from a `Provider` to just the string `namespace`? I'm wondering if we should return the full `Provider` instead.

cahrens: Or update the documentation to make it clear that it returns the provider namespaces.

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
aherbst-broad: NotRegistered

cahrens: Ah, I see. I was imagining it only returned things that they had registered at some point, but possibly things could be in an error state.

service/src/main/java/bio/terra/profile/service/azure/AzureService.java
aherbst-broad: Returning the namespace should be sufficient unless there is a compelling need for consumers of this method to have the full provider. I'll update the doc.

service/src/main/java/bio/terra/profile/service/profile/exception/MissingRequiredProvidersException.java
aherbst-broad: Done, see https://github.com/DataBiosphere/terra-billing-profile-manager/blob/45e2e4491708777d063fdbab60459b013909869d/service/src/main/resources/api/service_openapi.yaml#L65-L71

cahrens (approved):

PR #47, opened 2022-10-01
WOR-546 ignore offers from wrong publishers

dvoet:
I was wearing my hacker hat this weekend (alright, I usually wear it) and realized that I don't think we are picky enough in the Azure app deployments we trust. Anybody could create an offer and name it `terra-dev`. This PR adds another restriction that an Azure app deployment must be from a configured publisher.

https://broadworkbench.atlassian.net/browse/WOR-546

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -93,6 +93,7 @@ profile:
 
     application-offers:
       - name: terra-dev-preview
+        publisher: thebroadinstituteinc1615909626976
         authorized-user-key: authorizedTerraUser
 
     required-providers:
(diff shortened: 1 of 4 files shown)


service/src/main/resources/application.yml
```diff
@@ -93,6 +93,7 @@ profile:
 
     application-offers:
       - name: terra-dev-preview
+        publisher: thebroadinstituteinc1615909626976
```
dvoet, line 96: I found this value using a debug session, I don't know where else to find it.

aherbst-broad:
Thanks for doing this.

aherbst-broad (approved):

PR #49, opened 2022-10-04
[WOR-551] Add azure controller test

aherbst-broad:
While looking into smoke tests for BPM, I found that we 500 when no subscription ID is passed to the `getManagedApps` endpoint. This PR fixes that issue by adding a global exception handler to return 400 on MethodArgumentTypeMismatchException and MissingServletRequestParameterException, as well as a test for the controller.


cahrens:
Need to update Swagger documentation.

cahrens (approved):

blakery (approved):

PR #53, opened 2022-10-07
[WOR-560] Add test for the create profile flight

aherbst-broad:
This [sentry](https://sentry.io/organizations/broad-institute/issues/3653521049/?project=6608698&query=is%3Aunresolved) is caused by the [PR I merged here](https://github.com/DataBiosphere/terra-billing-profile-manager/pull/45/files). Specifically, the addition of the `getRequiredFoo` methods means Jackson will attempt to serialize those fields even for GCP profiles. 

We are lacking coverage for the `CreateProfileFlight`, so I've added that which gives us regression coverage on further serialization issues with this class and that flight in general.

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java
+++ b/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java
@@ -17,15 +17,16 @@ import java.util.List;
 record CreateProfileVerifyAccountStep(
     CrlService crlService, BillingProfile profile, AuthenticatedUserRequest user) implements Step {
 
+  public static List<String> PERMISSIONS_TO_TEST = List.of("billing.resourceAssociations.create");
+
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException {
     var billingCow = crlService.getBillingClientCow(user);
 
-    var permissionsToTest = List.of("billing.resourceAssociations.create");
     var testPermissionsRequest =
         TestIamPermissionsRequest.newBuilder()
-            .setResource(BillingAccountName.of(profile.billingAccountId().get()).toString())
-            .addAllPermissions(permissionsToTest)
+            .setResource(BillingAccountName.of(profile.getRequiredBillingAccountId()).toString())
+            .addAllPermissions(PERMISSIONS_TO_TEST)
             .build();
 
     final TestIamPermissionsResponse testPermissionsResponse;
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java
```diff
@@ -17,15 +17,16 @@
...
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException {
     var billingCow = crlService.getBillingClientCow(user);
 
-    var permissionsToTest = List.of("billing.resourceAssociations.create");
     var testPermissionsRequest =
         TestIamPermissionsRequest.newBuilder()
-            .setResource(BillingAccountName.of(profile.billingAccountId().get()).toString())
-            .addAllPermissions(permissionsToTest)
+            .setResource(BillingAccountName.of(profile.getRequiredBillingAccountId()).toString())
```
aherbst-broad, line 28: This change trues up the exception thrown here to be MissingRequiredFieldsException; previously it was throwing IllegalArgumentException due to de-referencing a null optional, which was different than the exception we are throwing on the similar Azure codepath.

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyDeployedApplicationStep.java
```diff
@@ -35,7 +35,7 @@ public StepResult doStep(FlightContext context) {
                           Objects.equals(
                                   app.getManagedResourceGroupId(),
                                   profile.getRequiredManagedResourceGroupId())
-                              && app.getSubscriptionId() == profile.getRequiredSubscriptionId())
+                              && Objects.equals(app.getSubscriptionId(), profile.getRequiredSubscriptionId()))
```
aherbst-broad, line 38: We were getting away with a bad object reference comparison; we want an bit-for-bit object value comparison instead. This is working today because I believe the getManagedAppDeployments method is just reflecting the UUID object back in it's return value.

service/src/main/java/bio/terra/profile/service/profile/model/BillingProfile.java
```diff
@@ -97,16 +99,24 @@ public ProfileModel toApiProfileModel() {
         .createdBy(createdBy);
   }
 
+  @JsonIgnore
```
aherbst-broad, line 102: This is the actual fix.

service/src/test/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlightTest.java
```diff
@@ -0,0 +1,203 @@
...
+  AuthenticatedUserRequest userRequest =
+      AuthenticatedUserRequest.builder()
+          .setToken("fake-token")
+          .setSubjectId("fake-sub")
+          .setEmail("example@example.com")
+          .build();
+
+  @Test
+  void createGcpProfileSuccess() {
+    var billingCow = mock(CloudBillingClientCow.class);
```
cahrens, line 50: Moo?

cahrens (approved):

marctalbott (approved):

PR #43, opened 2022-09-20
[WOR-520] Tag Azure MRG with billing profile ID

aherbst-broad:
## Context
([Related ticket)](https://broadworkbench.atlassian.net/browse/WOR-520)
We need to tag an MRG with a billing profile ID at the time of creation so Sam can use it when creating a pet managed identity. This came up during testing in a BEE, while attempting to add a VM to an MRG for a workspace.

## This PR
* Adds a new step to the azure create profile flight, which adds a tag to the MRG with the billing profile ID. 
* Adds related tests

PR #57, opened 2022-10-14
[WOR-156] Limit one billing profile per Azure managed app

marctalbott:
Ticket: [WOR-156](https://broadworkbench.atlassian.net/browse/WOR-156)
* add uniqueness constraint to prevent multiple billing profiles with the same managed app coordinates
* add optional query parameter to `listManagedApps` endpoint to specify whether to include managed apps that are already assigned to a billing profile

### DIFF
--- a/service/src/main/java/bio/terra/profile/db/ProfileDao.java
+++ b/service/src/main/java/bio/terra/profile/db/ProfileDao.java
@@ -97,7 +97,7 @@ public class ProfileDao {
           keyHolder.getString("created_by"));
     } catch (DuplicateKeyException ex) {
       if (ex.getMessage()
-          .contains("billing_profile_tenant_id_subscription_id_managed_resource__key")) {
+          .contains("billing_profile_subscription_id_managed_resource_group_id_key")) {
         throw new DuplicateManagedApplicationException("Managed application already in use.");
       } else {
         throw ex;
@@ -121,14 +121,13 @@ public class ProfileDao {
     return jdbcTemplate.query(SQL_LIST, params, new BillingProfileMapper());
   }
 
-  public List<String> listManagedResourceGroupsInSubscription(UUID tenantId, UUID subscriptionId) {
+  public List<String> listManagedResourceGroupsInSubscription(UUID subscriptionId) {
     var params =
         new MapSqlParameterSource()
-            .addValue("tenantId", tenantId)
             .addValue("subscriptionId", subscriptionId);
     return jdbcTemplate.queryForList(
         "SELECT managed_resource_group_id from billing_profile"
-            + " where tenant_id = :tenantId and subscription_id = :subscriptionId",
+            + " where subscription_id = :subscriptionId",
         params,
         String.class);
   }
--- a/service/src/main/java/bio/terra/profile/db/ProfileDao.java
+++ b/service/src/main/java/bio/terra/profile/db/ProfileDao.java
@@ -4,7 +4,6 @@ import bio.terra.common.db.ReadTransaction;
 import bio.terra.common.db.WriteTransaction;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.profile.model.CloudPlatform;
-import bio.terra.profile.service.azure.model.ManagedAppCoordinates;
 import bio.terra.profile.service.profile.exception.DuplicateManagedApplicationException;
 import bio.terra.profile.service.profile.exception.ProfileInUseException;
 import bio.terra.profile.service.profile.exception.ProfileNotFoundException;
@@ -29,6 +28,7 @@ import org.springframework.stereotype.Repository;
 
 @Repository
 public class ProfileDao {
+
   private final NamedParameterJdbcTemplate jdbcTemplate;
 
   // SQL select string constants
@@ -85,20 +85,21 @@ public class ProfileDao {
       jdbcTemplate.update(sql, params, keyHolder);
 
       return new BillingProfile(
-              profile.id(),
-              profile.displayName(),
-              profile.description(),
-              profile.biller(),
-              profile.cloudPlatform(),
-              profile.billingAccountId(),
-              profile.tenantId(),
-              profile.subscriptionId(),
-              profile.managedResourceGroupId(),
-              keyHolder.getInstant("created_date"),
-              keyHolder.getInstant("last_modified"),
-              keyHolder.getString("created_by"));
+          profile.id(),
+          profile.displayName(),
+          profile.description(),
+          profile.biller(),
+          profile.cloudPlatform(),
+          profile.billingAccountId(),
+          profile.tenantId(),
+          profile.subscriptionId(),
+          profile.managedResourceGroupId(),
+          keyHolder.getInstant("created_date"),
+          keyHolder.getInstant("last_modified"),
+          keyHolder.getString("created_by"));
     } catch (DuplicateKeyException ex) {
-      if (ex.getMessage().contains("billing_profile_tenant_id_subscription_id_managed_resource__key")) {
+      if (ex.getMessage()
+          .contains("billing_profile_tenant_id_subscription_id_managed_resource__key")) {
         throw new DuplicateManagedApplicationException("Managed application already in use.");
       } else {
         throw ex;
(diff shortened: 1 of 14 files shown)


service/src/main/java/bio/terra/profile/db/ProfileDao.java
```diff
@@ -76,21 +80,30 @@ public BillingProfile createBillingProfile(
...
+          profile.biller(),
+          profile.cloudPlatform(),
+          profile.billingAccountId(),
+          profile.tenantId(),
+          profile.subscriptionId(),
+          profile.managedResourceGroupId(),
+          keyHolder.getInstant("created_date"),
+          keyHolder.getInstant("last_modified"),
+          keyHolder.getString("created_by"));
+    } catch (DuplicateKeyException ex) {
```
marctalbott, line 99: I added this (and the dedicated exception) so that we could distinguish between a `DuplicateKeyException` because the billing profile id is already in use and one where the managed app is already in use. That said, I don't love having to make the determination based on the contents of the exception message. It will be pretty stable as the substring we're searching for is the name of the uniqueness constraint which is theoretically pretty stable. Still, if anyone has a better idea of how to do this, I'm all ears.

service/src/main/resources/application.yml
```diff
@@ -32,7 +32,7 @@ env:
 # contains the configuration of the BPM test application. We can use that application in our
 # integration testing to make sure the application code paths are working. However, we do not
 # want it to appear in production environments.
-spring.config.import: optional:file:service/build/resources/main/generated/local-properties.yaml
+spring.config.import: optional:file:build/resources/main/generated/local-properties.yaml
```
marctalbott, line 35: Spring checks the classpath to find this file. My classpath included `.../terra-billing-profile-manager/service/`, but not `.../terra-billing-profile-manager/`, so having `service/` at the beginning of this file was preventing Spring from importing the config correctly since it was looking in `.../terra-billing-profile-manager/service/service/build/...`. This meant the config rendered by `render_configs.sh` couldn't be found and my local BPM wasn't able to successfully call Azure since the client ids and secrets were wrong.

Not sure if this was just a bug or if it was specific to my machine. If a reviewer would be so kind as to try this out, I'd appreciate it!
steps to repro:
1. render configs `./render_configs.sh local`
2. run BPM
3. call the `listManagedApps` API - http://localhost:8080/swagger-ui.html#/Azure/getManagedAppDeployments

service/src/test/java/bio/terra/profile/service/azure/AzureServiceUnitTest.java
```diff
@@ -105,48 +107,130 @@ public void getManagedApps() {
...
+        .subscriptionId(subId)
+        .managedResourceGroupId(unassignedTerraAppManagedResourceGroupId)
+        .applicationDeploymentName(applicationName);
+
+    var includeAssignedResult = azureService.getAuthorizedManagedAppDeployments(subId, true, user);
+    assertEquals(List.of(assignedAzureManagedAppModel, unassignedAzureManagedAppModel),
+        includeAssignedResult);
+
+    appsList = Stream.of(assignedTerraApp, unassignedTerraApp);
+    when(appService.getApplicationsForSubscription(eq(subId))).thenReturn(appsList);
```
marctalbott, line 205: I had to introduce this to get around Streams only being readable one time. When I call `getAuthorizedManagedAppDeployments` for the second time below, the mock reuses the same `appsList` Stream it used the first time the method was called. However, that Stream has already been read, so it throws an exception. 

I don't like it, but I couldn't figure out how to get Streams to do what I wanted so this was my workaround. If anyone knows of a better way, let me know!

service/src/main/resources/db/changesets/20221011_unique_managed_application.yaml
```diff
@@ -0,0 +1,9 @@
+databaseChangeLog:
+  - changeSet:
+      id: unique_managed_application
+      author: mtalbott
+      changes:
+        - addUniqueConstraint:
```
aherbst-broad, line 6: Are there any existing profiles in dev that will violate this constraint and cause the migration to fail when applied?

service/src/main/resources/application.yml
cahrens: @aherbst-broad you inserted "service" here. Can you take a look?

https://github.com/DataBiosphere/terra-billing-profile-manager/pull/43/files#r976541391

service/src/main/java/bio/terra/profile/db/ProfileDao.java
```diff
@@ -109,6 +121,18 @@ public List<BillingProfile> listBillingProfiles(int offset, int limit, Collectio
     return jdbcTemplate.query(SQL_LIST, params, new BillingProfileMapper());
   }
 
+  public List<String> listManagedResourceGroupsInSubscription(UUID tenantId, UUID subscriptionId) {
```
cahrens, line 124: Is it necessary to limit to a specific tenant?

service/src/main/resources/db/changesets/20221011_unique_managed_application.yaml
cahrens: Should it just be unique between `subscription_id` and `managd_resource_group_id`?

service/src/main/resources/application.yml
aherbst-broad: I think the issue is intellij not using the correct working directory by default when running the app. Maybe we can create a default Intellij configuration similar to WSM to avoid this. Anyway, I think this change is fine, I can run successfully using the new path.

service/src/main/resources/db/changesets/20221011_unique_managed_application.yaml
marctalbott: > Are there any existing profiles in dev that will violate this constraint and cause the migration to fail when applied?

Hadn't checked, but yes there are a number of existing profiles that will violate this constraint. I believe I can get around that by making the constraint deferrable, I'll play around with that.

marctalbott: > Should it just be unique between `subscription_id` and `managd_resource_group_id`?

I went with all three because they represent the full managed app coordinates, but I just checked and Azure subscription IDs are globally unique, so yes we can make this unique just between the subscription and managed resource group id.

service/src/main/java/bio/terra/profile/db/ProfileDao.java
marctalbott: It is not, I thought it was possible for subscriptions in different tenants to share the same ID, but subscription IDs are globally unique, so I'll remove the tenant ID here and there's a few test checks that can be modified/removed as well

service/src/main/resources/db/changesets/20221011_unique_managed_application.yaml
cahrens: Many of the problematic profiles are probably mine. If it becomes a problem, I am fine deleting mine.

marctalbott: constraint deferral did not do the trick, looks like the only option is to delete the offending billing profiles. Fortunately, all of the offending profiles have been created by our team. I'll post the list in Slack and make sure everyone's okay with me deleting duplicates

service/src/main/resources/db/changesets/20221011_unique_managed_application.yaml
```diff
@@ -0,0 +1,9 @@
...
+            constraint_name: unique_managed_application
+            tableName: billing_profile
+            columnNames: subscription_id, managed_resource_group_id
```
cahrens, line 9: Now I'm getting confused… an MRG can only be linked to a single billing profile. So why is it the combination of subscription_id and MRG that is unique? Wouldn't it be the combination of the primary key and `managed_resource_group_id`?

marctalbott: The unique constraint will make sure there is only one record with a given subscription/MRG combo. Since a subscription/MRG combo represents a single managed app deployment, this constraint will trigger and throw an exception if we try to create a second record using the same subscription/MRG combo (managed app deployment)

cahrens: I was thinking that the `managed_resource_group_id` uniquely determined a MRG deployment… is that not the case?

marctalbott: Hmm, I don't know. I thought that `managed_resource_group_id` was unique within a subscription (or maybe a tenant?), but that it wasn't necessarily globally unique. Azure's namespaces confuse me. If it is globally unique, then we can make just the `managed_resource_group_id` column unique.

marctalbott: This link makes me think that the `managed_resource_group_id` isn't globally unique. I'm not sure if we're dealing with `deployment` or `resourcegroups`, but neither have a global naming scope

cahrens (approved):
LGTM once the Swagger description string is updated.

blakery (approved):

PR #89, opened 2022-11-28
Add check for Microsoft.Relay (WOR-617).

cahrens:
Still trying to get to the point where I can manually test this, but the change is straightforward.


marctalbott (approved):

aherbst-broad (approved):
lgtm; thinking about this a bit more, you should be able to test this against the workspaces sub by unregistering the Microsoft.Relay provider and observing a failure when creating a billing profile against an MRG in that sub.

PR #91, opened 2022-11-29
WOR-634

blakery:
[WOR-634](https://broadworkbench.atlassian.net/browse/WOR-634)

* Adds a step to unlink the MRG in the billing profile delete flight 
* Unlink MRG in the undo step of linking an MRG

On the develop branch, when I create a billing profile, delete it, and then try to create a new billing profile re-using the same MRG, I get a 409 back with the error message: `"MRG has existing tag [mrg_id = ..., existing tag = terra.billingProfileId, value =...]"`

After the changes, the above process succeeds.

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
+++ b/service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
@@ -78,12 +78,14 @@ public class ApplicationService {
    * @param tag Tag name
    * @throws MissingRequiredFieldsException if the tag is not present.
    */
-  public void removeTagFromMrg(UUID tenantId, UUID subscriptionId, String managedResourceGroupId, String tag) {
+  public void removeTagFromMrg(
+      UUID tenantId, UUID subscriptionId, String managedResourceGroupId, String tag) {
     var mrgResource = crlService.getResourceGroup(tenantId, subscriptionId, managedResourceGroupId);
     if (!mrgResource.tags().containsKey(tag)) {
       throw new MissingRequiredFieldsException(
-          String.format("Cannot delete missing tag from MRG [mrg_id = %s, tag = %s]", managedResourceGroupId, tag)
-      );
+          String.format(
+              "Cannot delete missing tag from MRG [mrg_id = %s, tag = %s]",
+              managedResourceGroupId, tag));
     }
 
     // dupe existing tags to a mutable hashmap
@@ -91,6 +93,4 @@ public class ApplicationService {
     tags.remove(tag);
     crlService.updateTagsForResource(tenantId, subscriptionId, mrgResource.id(), tags);
   }
-
-
 }
--- a/service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
+++ b/service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
@@ -2,7 +2,6 @@ package bio.terra.profile.service.azure;
 
 import bio.terra.profile.service.crl.CrlService;
 import bio.terra.profile.service.profile.exception.DuplicateTagException;
-import bio.terra.profile.service.profile.exception.MissingRequiredFieldsException;
 import com.azure.resourcemanager.managedapplications.models.Application;
 import java.util.HashMap;
 import java.util.UUID;
@@ -76,7 +75,6 @@ public class ApplicationService {
    * @param subscriptionId Subscription ID associated with the MRG
    * @param managedResourceGroupId ID for the MRG
    * @param tag Tag name
-   * @throws MissingRequiredFieldsException if the tag is not present.
    */
   public void removeTagFromMrg(
       UUID tenantId, UUID subscriptionId, String managedResourceGroupId, String tag) {
--- a/service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
+++ b/service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
@@ -2,6 +2,7 @@ package bio.terra.profile.service.azure;
 
 import bio.terra.profile.service.crl.CrlService;
 import bio.terra.profile.service.profile.exception.DuplicateTagException;
+import bio.terra.profile.service.profile.exception.MissingRequiredFieldsException;
 import com.azure.resourcemanager.managedapplications.models.Application;
 import java.util.HashMap;
 import java.util.UUID;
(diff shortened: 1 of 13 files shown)


service/src/main/java/bio/terra/profile/service/profile/flight/delete/UnlinkBillingProfileIdFromMrgStep.java
```diff
@@ -0,0 +1,35 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import bio.terra.stairway.exception.RetryException;
+
+/** Removes the tag that associates a billing profile ID with an azure managed resource group */
+public record UnlinkBillingProfileIdFromMrgStep(
+    ApplicationService applicationService, BillingProfile profile) implements Step {
+  private static final String BILLING_PROFILE_ID_TAG = "terra.billingProfileId";
```
blakery, line 14: This constant is also in `LinkBillingProfileIdFromMrgStep`, so it should probably be located somewhere else. I'm not sure where the best place to put it is, in this project.

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
```diff
@@ -83,6 +84,10 @@ public void deleteProfile(UUID id, AuthenticatedUserRequest user) {
             .userRequest(user)
             .addParameter(ProfileMapKeys.PROFILE_ID, id)
             .addParameter(JobMapKeys.CLOUD_PLATFORM.getKeyName(), platform.name());
+    if (CloudPlatform.AZURE.equals(billingProfile.cloudPlatform())) {
```
blakery, line 87: It seemed silly to add all the individual app coordinates separately.
It also seems (a little) silly to keep passing the id separately, and then to only add the profile when we need the app coordinates.
I also didn't want to rely on retrieving the profile from the DAO later on, since we're deleting it.

Should I just get rid of the profile id param and just use the profile? I think that would be cleaner.

service/src/main/java/bio/terra/profile/service/profile/flight/delete/DeleteProfileFlight.java
```diff
@@ -19,8 +20,8 @@ public DeleteProfileFlight(FlightMap inputParameters, Object applicationContext)
 
     ApplicationContext appContext = (ApplicationContext) applicationContext;
     ProfileDao profileDao = appContext.getBean(ProfileDao.class);
-    CrlService crlService = appContext.getBean(CrlService.class);
```
blakery, line 22: This was unused

service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
```diff
@@ -67,4 +68,29 @@ public void addTagToMrg(
...
+   * @param subscriptionId Subscription ID associated with the MRG
+   * @param managedResourceGroupId ID for the MRG
+   * @param tag Tag name
+   * @throws MissingRequiredFieldsException if the tag is not present.
+   */
+  public void removeTagFromMrg(
+      UUID tenantId, UUID subscriptionId, String managedResourceGroupId, String tag) {
+    var mrgResource = crlService.getResourceGroup(tenantId, subscriptionId, managedResourceGroupId);
+    if (!mrgResource.tags().containsKey(tag)) {
+      throw new MissingRequiredFieldsException(
```
blakery, line 85: Does this exception seem reasonable? Or should I create a more specific one?

cahrens: I'm wondering if we really need to throw an exception in this case. If the tab isn't present in the MRG, then calling this is just a no-op. While perhaps it is unexpected, the end goal is the same (the tag isn't present).

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
cahrens: I don't think this code needs to be specific to Azure.

I think it makes sense to remove the profile_id parameter and just always add the profile itself.

service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
blakery: That's a good point. Basically framing the operation as "make sure this tag isn't there".

aherbst-broad: +1 to no exception

service/src/main/java/bio/terra/profile/service/profile/flight/delete/UnlinkBillingProfileIdFromMrgStep.java
aherbst-broad: I suggest a common constants class.

aherbst-broad (approved):

service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
```diff
@@ -67,4 +68,26 @@ public void addTagToMrg(
...
   }
+
+  /**
+   * Removes a tag to the MRG associated with a given profile.
+   *
+   * @param tenantId Tenant ID associated with the MRG
+   * @param subscriptionId Subscription ID associated with the MRG
+   * @param managedResourceGroupId ID for the MRG
+   * @param tag Tag name
+   * @throws MissingRequiredFieldsException if the tag is not present.
```
cahrens, line 79: Need to update this since it doesn't throw an exception.

service/src/main/java/bio/terra/profile/service/azure/ApplicationService.java
```diff
@@ -67,4 +68,26 @@ public void addTagToMrg(
...
+  public void removeTagFromMrg(
+      UUID tenantId, UUID subscriptionId, String managedResourceGroupId, String tag) {
+    var mrgResource = crlService.getResourceGroup(tenantId, subscriptionId, managedResourceGroupId);
+    if (mrgResource.tags().containsKey(tag)) {
+      // dupe existing tags to a mutable hashmap
+      HashMap<String, String> tags = new HashMap<>(mrgResource.tags());
+      tags.remove(tag);
+      crlService.updateTagsForResource(tenantId, subscriptionId, mrgResource.id(), tags);
+    }
+
```
cahrens, line 90: Extra blank lines

service/src/main/java/bio/terra/profile/service/profile/flight/create/LinkBillingProfileIdToMrgStep.java
```diff
@@ -31,6 +31,14 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
   public StepResult undoStep(FlightContext context) throws InterruptedException {
-    return StepResult.getStepResultSuccess();
+    try {
+      var tenantId = profile.getRequiredTenantId();
+      var subscriptionId = profile.getRequiredSubscriptionId();
+      var mrgId = profile.getRequiredManagedResourceGroupId();
+      applicationService.removeTagFromMrg(tenantId, subscriptionId, mrgId, BILLING_PROFILE_ID_TAG);
+      return StepResult.getStepResultSuccess();
+    } catch (Exception ex) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
```
cahrens, line 41: I think you can simplify this now and remove the try/catch

service/src/main/java/bio/terra/profile/service/profile/flight/delete/UnlinkBillingProfileIdFromMrgStep.java
```diff
@@ -0,0 +1,35 @@
...
+    try {
+      var tenantId = profile.getRequiredTenantId();
+      var subscriptionId = profile.getRequiredSubscriptionId();
+      var mrgId = profile.getRequiredManagedResourceGroupId();
+
+      applicationService.removeTagFromMrg(tenantId, subscriptionId, mrgId, BILLING_PROFILE_ID_TAG);
+
+      return StepResult.getStepResultSuccess();
+    } catch (Exception ex) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
```
cahrens, line 27: Same question here-- do we need the explicit try/catch?

service/src/main/java/bio/terra/profile/service/profile/flight/delete/UnlinkBillingProfileIdFromMrgStep.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
cahrens, line 33: Should undoStep add the tag back?

blakery:
[TOAZ-22](https://broadworkbench.atlassian.net/browse/TOAZ-224) landed quicker than anticipated, so the tags on the MRG will be imminently obsolete.

PR #93, opened 2022-12-01
[WOR-637] Store MRG coordinates in Sam at the time of profile creation

aherbst-broad:


service/src/test/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlightTest.java
```diff
@@ -179,25 +144,37 @@ void createAzureProfileSuccess() {
...
+        .thenReturn(azureConfiguration.getRequiredProviders());
+    doThrow(SamExceptionFactory.create("foo", new InterruptedException()))
+        .when(samService)
+        .createManagedResourceGroup(any(), eq(userRequest));
+    var profile = ProfileFixtures.createAzureBillingProfile(tenantId, subId, mrgId);
+
+    assertThrows(
+        SamInterruptedException.class, () -> profileService.createProfile(profile, userRequest));
+
+    verify(samService).deleteManagedResourceGroup(profile.id(), userRequest);
```
cahrens, line 178: This is interesting-- `createManagedResourceGroup` is what threw the exception, so I wouldn't necessarily expect `deleteManagedResourceGroup` to succeed (I'm imagining the resource group would not be have been created). But I suppose the NOT_FOUND handling catches that case.

service/src/main/java/bio/terra/profile/service/iam/SamService.java
```diff
@@ -345,6 +352,71 @@ public void deleteProfileResource(AuthenticatedUserRequest userRequest, UUID pro
...
+    } catch (ApiException e) {
+      if (e.getCode() == HttpStatus.CONFLICT.value()) {
+        // MRG has already been created in Sam, move on
+        logger.info(
+            "Mrg record already exists in Sam for profile {},  mrg id = {}, tenant = {}, subscription = {}",
+            profile.id(),
+            profile.managedResourceGroupId(),
+            profile.tenantId(),
+            profile.subscriptionId());
+        return;
```
blakery, line 413: Is `CONFLICT` returned when there's a record in Sam where the profile.id is there, but the MRG coordinates don't match, or vice versa? 
I would think we'd want to throw the exception unless the record in Sam matches exactly (in which case this is essentially a no-op), and otherwise return an error that the MRG is already assigned.

blakery (approved):

PR #102, opened 2022-12-08
WOR-603

blakery:
update to latest versions of sam client and common-lib

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -19,7 +19,7 @@ dependencies {
 		}
 		because 'previous versions have a security vulnerability - this is a direct dependency to force the version to override opencensus'
 	}
-	implementation 'bio.terra:terra-common-lib:0.0.68-SNAPSHOT'
+	implementation 'bio.terra:terra-common-lib:0.0.77-SNAPSHOT'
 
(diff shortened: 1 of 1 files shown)


marctalbott:
Did you confirm that this upgrades the Scala version?

aherbst-broad (approved):

cahrens (approved):

cahrens:
> Did you confirm that this upgrades the Scala version?

@blakery bumping Marc's question, as I didn't see it before I approved the PR.

blakery:
> Did you confirm that this upgrades the Scala version?

There's a bit of ambiguity in the dependency tree, so I'm not _certain_. 
In the compileClasspath, it's resolving to 2.13.5, despite being on the latest sam client.
In the productionRuntimeClasspath, it lists 2.13.5 as the requested version, along with an older sam client, but resolves to 2.13.10, along with the latest sam client.
Then later in the productionRuntimeClasspath, it lists 2.13.10 both an initial and resolved version.

So: I _think_ it upgrades the version. Some of what I'm seeing might also be some sort of caching on my system. I've cleaned, rebuilt, and reimported the project multiple times, but I guess I haven't cleared my gradle cache recently, now that I think about it.

PR #115, opened 2023-01-03
WOR-589

blakery:
[WOR-589](https://broadworkbench.atlassian.net/browse/WOR-589)
This PR is for some initial refactoring of tests into true unit tests.

The next step I'm working on is getting a test context for spring to work without the DB, for integration tests.

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java
+++ b/service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java
@@ -14,7 +14,7 @@ import com.google.iam.v1.TestIamPermissionsResponse;
 import java.util.List;
 
 /** Step to verify the user has access to a GCP profile's billing account. */
-record CreateProfileVerifyAccountStep(
+public record CreateProfileVerifyAccountStep(
     CrlService crlService, BillingProfile profile, AuthenticatedUserRequest user) implements Step {
 
   public static List<String> PERMISSIONS_TO_TEST = List.of("billing.resourceAssociations.create");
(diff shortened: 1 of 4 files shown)


service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileVerifyAccountStep.java
```diff
@@ -14,7 +14,7 @@
 import java.util.List;
 
 /** Step to verify the user has access to a GCP profile's billing account. */
-record CreateProfileVerifyAccountStep(
+public record CreateProfileVerifyAccountStep(
```
blakery, line 17: This needed to be public instead of package scoped, in order to set up tests.

service/src/test/java/bio/terra/profile/common/MetricsUtilsUnitTest.java
```diff
@@ -0,0 +1,55 @@
...
+  void tearDown() {
+    meterRegistry.clear();
+    Metrics.globalRegistry.clear();
+  }
+
+  @Test
+  void createGcpProfileMetrics() {
+    var profile = ProfileFixtures.createGcpBillingProfile("ABCD1234");
+
+    MetricUtils.recordProfileCreation(() -> profile, profile.cloudPlatform());
```
cahrens, line 35: By doing the test separately, we have no coverage that we actually call `recordProfileCreation` as part of profile creation. Can you spy on that call in `CreateProfileFlightTest` to also ensure the method is called?

marctalbott (approved):

cahrens (approved):

PR #116, opened 2023-01-03
Try preventing major upgrades to spring-boot (WOR-680).

cahrens:
My first attempt in https://github.com/DataBiosphere/terra-billing-profile-manager/pull/95 did not work (see https://github.com/DataBiosphere/terra-billing-profile-manager/pull/111). Trying a different syntax.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,4 +15,4 @@ updates:
       prefix: "[No Ticket]"
     ignore:
       - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"
-        versions: ["3.x"]
+        update-types: ["version-update:semver-major"]


.github/dependabot.yml
```diff
@@ -15,4 +15,4 @@ updates:
       prefix: "[No Ticket]"
     ignore:
       - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"
-        versions: ["3.x"]
+        update-types: ["version-update:semver-major"]
```
cahrens, line 18: Syntax from https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

cahrens:
Note that I also added the suggested "ignore comment" to https://github.com/DataBiosphere/terra-billing-profile-manager/pull/111, so having this in our dependabot config file may be unnecessary. But I think the documentation is helpful.

aherbst-broad (approved):

blakery (approved):

PR #146, opened 2023-02-14
[WOR-507] Set no-store Cache-Control header on all authenticated APIs

marctalbott:
Ticket: [WOR-507](https://broadworkbench.atlassian.net/browse/WOR-507)
* set cache-control header to `no-store` so that browsers will not cache the returned content. This header is set only on `/api/*` endpoints as it is not needed for unauthenticated requests

Before:
```
curl -X GET 'http://localhost:8080/api/profiles/v1?offset=0&limit=10' --header "Authorization: Bearer `gcloud auth print-access-token marc.testbott@gmail.com`" -i
HTTP/1.1 200 
X-Request-ID: axzk5MzR
Content-Type: application/json
Content-Length: 481
```

After:
```
curl -X GET 'http://localhost:8080/api/profiles/v1?offset=0&limit=10' --header "Authorization: Bearer `gcloud auth print-access-token marc.testbott@gmail.com`" -i                                                            
HTTP/1.1 200 
Cache-Control: no-store
X-Request-ID: ey3KL2Xj
Content-Type: application/json
Content-Length: 481
```

Note that the reason there are fewer headers shown here is because I grabbed the output from a locally running BPM and additional headers (like you may see in the ticket description) are added when it's running in k8s by the proxy presumably

[WOR-507]: https://broadworkbench.atlassian.net/browse/WOR-507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/Main.java
+++ b/service/src/main/java/bio/terra/profile/app/Main.java
@@ -33,9 +33,9 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
       "bio.terra.profile"
     })
 @ConfigurationPropertiesScan(basePackages = {"bio.terra.profile"})
+@ServletComponentScan(basePackages = {"bio.terra.profile"})
 @EnableRetry
 @EnableTransactionManagement
-@ServletComponentScan
 public class Main {
   public static void main(String[] args) {
     new SpringApplicationBuilder(Main.class).initializers(new LoggingInitializer()).run(args);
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/profile/app/Main.java
```diff
@@ -34,6 +35,7 @@
 @ConfigurationPropertiesScan(basePackages = {"bio.terra.profile"})
 @EnableRetry
 @EnableTransactionManagement
+@ServletComponentScan
```
cahrens, line 38: What is this?

marctalbott: It tells Spring to look for Servlet components. The `@WebFilter` component isn't found by the normal `ComponentScan`, so this is required for Spring to find the new filter. Without it, BPM won't add the `Cache-Control` header. 

There's a little more info here - https://www.baeldung.com/spring-servletcomponentscan#using-servletcomponentscan-in-spring-boot

cahrens (approved):

PR #155, opened 2023-02-23
[WOR-714] Bump cloud-resource-library version

marctalbott:
Ticket: [WOR-714](https://broadworkbench.atlassian.net/browse/WOR-714)

- Tested by attempting to create a GCP billing profile with two different users, one with access to the Google billing account and one without. `CreateProfileVerifyAccountStep` (what uses the GCP billing CRL library) failed as expected for the user without access and succeeded for the one that did. 

[WOR-714]: https://broadworkbench.atlassian.net/browse/WOR-714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -37,11 +37,7 @@ dependencies {
 	implementation group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.2"
 	implementation group: "com.azure.resourcemanager", name: "azure-resourcemanager-subscription", version: "1.0.0-beta.2"
 
-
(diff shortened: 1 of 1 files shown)


cahrens (approved):
I'm confused as to why the code that was never merged is not needed. But given that we don't currently use the GCP code in BPM I figure we can fix any issues that arise down the line, and it's best to get this dependency up-to-date ASAP.

blakery (approved):

PR #168, opened 2023-03-21
WOR-550 - BPM contract testing

blakery:
[WOR-550](https://broadworkbench.atlassian.net/browse/WOR-550)
Add single pact test that can generate a pact file when run manually

[WOR-550]: https://broadworkbench.atlassian.net/browse/WOR-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pact/src/main/java/bio/terra/profile/consumer/SamStatusApiPactTest.java
+++ b/pact/src/main/java/bio/terra/profile/consumer/SamStatusApiPactTest.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 public class SamStatusApiPactTest extends ConsumerPactTest {
 
 
-  @Pact(consumer = "bpm-consumer")
   @Override
   protected RequestResponsePact createPact(PactDslWithProvider builder) {
     return builder


pact/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+    id 'bio.terra.profile.java-spring-conventions'
+    id "au.com.dius.pact" version "4.3.10"
+}
+
+
+dependencies {
+    implementation 'au.com.dius.pact.consumer:junit:4.2.9'
+
+    implementation group: "org.broadinstitute.dsde.workbench", name: "sam-client_2.13", version: "0.1-e0d8da8"
+//    "au.com.dius:pact-jvm-provider-gradle:4.0.3"
```
cahrens, line 16: Is this comment meaningful?

pact/src/main/java/bio/terra/profile/consumer/SamStatusApiPactTest.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * The only way I've been able to get this to generate a pact successfully so far,
+ * has been running it directly in my IDE
+ * If I put it in a test folder, it doesn't get detected as containing any tests
+ * In this folder, the issue is that it isn't set up with a 'main' method
+ */
+public class SamStatusApiPactTest extends ConsumerPactTest {
+
+
+  @Pact(consumer = "bpm-consumer", provider = "sam-provider")
```
cahrens, line 25: Why does it need this annotation given that there are also the overrides at lines 38 and 43?

blakery: The consumer is required by the annotation, for some reason.
The provider, it turns out, is not - so I removed it.

There is another definition pattern that's a shorthand for this one, where both are needed (I couldn't get that one to work)

pact/src/main/java/bio/terra/profile/consumer/SamStatusApiPactTest.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * The only way I've been able to get this to generate a pact successfully so far,
+ * has been running it directly in my IDE
+ * If I put it in a test folder, it doesn't get detected as containing any tests
+ * In this folder, the issue is that it isn't set up with a 'main' method
+ */
+public class SamStatusApiPactTest extends ConsumerPactTest {
+
+
+  @Pact(consumer = "bpm-consumer")
```
cahrens, line 25: I can also delete the consumer annotation and it still generates the pact.

blakery: that worked for me, too
I wonder why some examples have both (I just double checked to make sure I wasn't miss-remembering)

cahrens (approved):

pact/src/main/java/bio/terra/profile/consumer/SamStatusApiPactTest.java
cahrens: I don't know… maybe we will figure out more as we get into more complicated usage. Have the annotation just seemed to duplicate the methods (though everything I say should be taken with a huge grain of salt because I haven't read about Pact yet!).

blakery: Figured it out:
The 'official' example has the annotation on the alternate usage. 
The example that had both was an unofficial example (although it was linked from the official page).
So the example was confused.

aherbst-broad (approved):

PR #179, opened 2023-03-31
WOR-835 - automate pact - rebase

blakery:
[WOR-835](https://broadworkbench.atlassian.net/browse/WOR-835)

In this PR:
* move pact tests to main project 
* refactor pact tests to use junit5 runner (this was the primary issue in running them in the main project before)
* create a gradle task so pact generation can be invoked explicitly by the github workflow
* add a github workflow to publish the pact

[WOR-835]: https://broadworkbench.atlassian.net/browse/WOR-835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/consumer_contract_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+#
+# https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2660368406/Getting+Started+with+Pact+Contract+Testing
+#
+#
+#
+# NOTE: The publish-contracts workflow will use the latest commit of the branch that triggers this workflow to publish the unique consumer contract version to Pact Broker.
+
+on:
+  pull_request:
+    branches: [ '**' ]
```
cahrens, line 15: I'm not clear about this syntax-- does it mean that it ignores branches?

service/src/test/java/bio/terra/profile/service/iam/SamServiceTest.java
marctalbott, line 1: It seems to me like it'd be valuable to keep our contract tests separate from regular unit tests, but still in the main project. What do you think?

.github/workflows/consumer_contract_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+        id: encode-pact
+        run: |
+          NON_BREAKING_B64=$(cat service/build/pacts/bpm-consumer-sam-provider.json | base64 -w 0)
+          echo "pact-b64=${NON_BREAKING_B64}" >> $GITHUB_OUTPUT
+  publish-contracts:
+    runs-on: ubuntu-latest
+    needs: [init-github-context, bpm-consumer-contract-tests]
+    steps:
+      - name: Dispatch to terra-github-workflows
+        uses: aurelien-baudet/workflow-dispatch@93e95b157d791ae7f42aef8f8a0d3d723eba1c31 #commit sha for v2.1.1
```
marctalbott, line 76: There's a `broadinstitute` version of `workflow-dispatch`, can we use that instead? I don't remember the differences but I tend to favor using in house versions for GHA

.github/workflows/consumer_contract_tests.yml
```diff
@@ -0,0 +1,82 @@
...
+# NOTE: The publish-contracts workflow will use the latest commit of the branch that triggers this workflow to publish the unique consumer contract version to Pact Broker.
+
+on:
+  pull_request:
+    branches: [ '**' ]
+    paths-ignore: [ '**.md' ]
+  push:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+  merge_group:
```
marctalbott, line 20: Don't know that we need `merge_group`, but I suppose it's fine in case we decide to start using a merge queue in the future

.github/workflows/consumer_contract_tests.yml
marctalbott: iirc it means this will run on pull requests against all branches. I'm also curious about whether we need to clean up PR pacts

blakery: Pacts are published [here](https://pact-broker.dsp-eng-tools.broadinstitute.org)

TBH, I'm not sure if we need to clean up PR pacts either.
They're indexed by commit hash, so once it's merged, it won't generate a separate pact.
It's also pretty compact in the grand scheme of things, and they can be removed manually; so I'm not especially concerned with it in the short term, since conventions are still being established.

It's plausible we should restrict it to PRs to `main`, though (not that it's a huge difference, since most of our PRs are against `main`).

service/src/test/java/bio/terra/profile/service/iam/SamServiceTest.java
blakery: I put them in with the regular tests because I was running it against our actual code, and letting our real usage determine what it needs.
In this scenario, the pacts are effectively mocks for actual tests, to some extent.

Also, we have the tags to distinguish the tests, and we already have unit tests in the same folders as what are effectively integration tests (the spring tests require a db).

I'm not necessarily set against moving them to a separate folder, but maybe a naming convention might work better?

.github/workflows/consumer_contract_tests.yml
blakery: I just ripped it off of what leo was using, so I don't know what the differences are either. I'm not attached to this one, though.

ichengchang: The leo consumer test yaml has been updated to the `broadinstitute` version as @marctalbott mentioned. Here's the latest code [https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/consumer_contract_tests.yml](https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/consumer_contract_tests.yml).

Just follow the new code and you'll be all set.

.github/workflows/consumer_contract_tests.yml
ichengchang: There is no need to clean up PR pacts on the Pact Broker. As @blakery said, normally we restrict PRs to trigger on `main`. When you push code to PR branch, the contract tests will run. Upon successful test run, this workflow will publish the contract to Pact Broker tagged with the latest commit, which is the version the Pact Broker cares about when calculating contract changes. Now the key part is, when your team decides everything looks good and merge PR to `main`, the same contract tests will run again which shouldn't cause no surprise, the contract will be published but tagged with the merge commit. Your repo HEAD is now tagged with merge commit, it's the same in Pact Broker. Furthermore, DevOps will be able to grab the latest commit for releases. Everything is in sync.

service/src/test/java/bio/terra/profile/service/iam/SamServiceTest.java
ichengchang: I don't have strong opinions on where contract tests code should live. However, I have a preference on using a separate module for pact code. In my case, I am working with Scala, and it gives me better control on what dependencies to include in case pact libraries conflicts with libraries used in the parent module.

.github/workflows/consumer_contract_tests.yml
```diff
@@ -0,0 +1,100 @@
...
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+
+jobs:
+  init-github-context:
+    runs-on: ubuntu-latest
+    outputs:
+      repo-branch: ${{ steps.extract-branch.outputs.repo-branch }}
```
ichengchang, line 28: We also need to add `repo-version` to the `outputs` parameter here. Follow the example in [https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/consumer_contract_tests.yml#L53](https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/consumer_contract_tests.yml#L53)

cahrens (approved):
Take my approval with a grain of salt-- I don't know much about this. I would trust @ichengchang much more!

.github/workflows/consumer_contract_tests.yml
```diff
@@ -0,0 +1,101 @@
...
+      repo-version: ${{ steps.extract-branch.outputs.repo-version }}
+
+    steps:
+      - uses: actions/checkout@v3
+      - id: extract-branch
+        run: |
+          GITHUB_EVENT_NAME=${{ github.event_name }}
+          if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
+            GITHUB_REF=${{ github.ref }}
```
ichengchang, line 37: Please include `GITHUB_SHA=${{ github.sha }}`. 

See [https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/consumer_contract_tests.yml#L66](https://github.com/DataBiosphere/leonardo/blob/develop/.github/workflows/consumer_contract_tests.yml#L66)

ichengchang (approved):
Looks good! You can disregard can-i-deploy for the moment as bpm-consumer hasn't been plugged into the Pact Broker CI / CD at the moment.

knapii-developments (approved):

PR #181, opened 2023-04-04
WOR-878 429 when pulling Azure spend data

sergiygetlin:
https://broadworkbench.atlassian.net/browse/WOR-878
Cache spend report results to minimize request to Azure


service/src/main/java/bio/terra/profile/service/spendreporting/azure/AzureSpendReportingService.java
```diff
@@ -147,4 +161,20 @@ private String getK8sNodeResourceGroupName(
         k8sResource.name(),
         k8sResource.regionName());
   }
+
+  @Scheduled(
```
aherbst-broad, line 165: @sergiygetlin is it possible to introduce a TTL per cache entry rather than flushing the cache in bulk every N minutes? I'm envisioning a scenario where someone hits this API, the cache is populated then hits it again a minute later after the cache is flushed and we make another call to azure.

sergiygetlin: @aherbst-broad There is a different way to remove items from cache like for instance "@CacheEvict" attribute, but it doesn't allow to setup TTL. Such thing as TTL should be set directly via Cache provider (Redis, etc.) since Spring provides only high-level cache abstraction to work with different cache solutions (at the same time it provides simple cache implementation such as ConcurrentHashMap). So, to use more sophisticated cache eviction policies we need to switch to non-default cache provider.

aherbst-broad (approved):
lgtm, let's try it out.

service/src/main/java/bio/terra/profile/service/spendreporting/azure/AzureSpendReportingService.java
```diff
@@ -51,6 +58,7 @@ public AzureSpendReportingService(
    * @param to End of the billing period
    * @return Spend data
    */
+  @Cacheable(value = CacheConfiguration.AZURE_SPEND_REPORT_CACHE_NAME)
   public SpendData getBillingProfileSpendData(
```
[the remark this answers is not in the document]
sergiygetlin: There are no Sam checks downstream, only communication with Azure.

cahrens (approved):
This is really cool!

PR #184, opened 2023-04-05
[WOR-892] Do not dispatch tag if this is a version bump

aherbst-broad:
We erroneously fire off a tag+publish+deploy on version bumps, which means we are attempting to tag+publish+deploy the same version twice which [results](https://github.com/DataBiosphere/terra-billing-profile-manager/actions/workflows/publish.yml) in a 409 in sherlock. This PR adds a check for version bump merges.

### DIFF
--- a/.github/actions/bump-skip/action.yml
+++ b/.github/actions/bump-skip/action.yml
@@ -1,0 +1,28 @@
+# This action must be done after the checkout action
+name: 'bump-skip'
+description: 'Set skip-out when we are doing a version bump'
+author: 'dd'
+inputs:
+  event-name:
+    description: 'github.event_name from the calling workflow'
+    required: true
+outputs:
+  is-bump:
+    description: 'yes if this is a push made by bumper; no if it is a regular push'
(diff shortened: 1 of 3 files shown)


.github/actions/bump-skip/action.yml
```diff
@@ -0,0 +1,28 @@
+# This action must be done after the checkout action
```
aherbst-broad, line 1: Stolen from the landing zone service.

.github/workflows/build-and-test.yml
```diff
@@ -86,6 +87,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        # Needed by sonar to get the git history for the branch the PR will be merged into.
+        with:
+          fetch-depth: 0
```
aherbst-broad, line 92: Sonarqube was erroring out like [this](https://github.com/DataBiosphere/terra-billing-profile-manager/actions/runs/4621998118/jobs/8174213641#step:9:38) on PRs until I added this.

marctalbott (approved):

sergiygetlin (approved):

PR #203, opened 2023-05-04
WOR-937 refactor Azure and GCP service calls 

blakery:
[WOR-937](https://broadworkbench.atlassian.net/browse/WOR-937)

* Splits CrlService into two separate classes, one for Azure resources, and one for GCP
* Make the resulting `AzureCrlService` and `GcpCrlService` just provide the initial resource objects 
  * It's possible this isn't the best approach, but it's at least consistent, now - before it both providing the raw resources, and providing some functionality over them
* Removes the `ApplicationService`, since this was inconsistently used and mostly acting as a pass through to functionality in `AzureCrlService`
  * I think that maybe this was originally intended as a more abstracted layer over the resources provided in `CrlService`, but the usage pattern had grown inconsistent over time, and I thought it would probably be better to make the clean reorganization as is first, then figure out what, if anything was needed as an intermediary layer

Previously, the dependency structure looked something like:
`AzureService` => `ApplicationService`, `CrlService`
`ApplicationService` => `CrlService`
Where azure service contained a mix of calls delegating to `ApplicationService` and using resources obtained from `CrlService` directly.

After these changes, `AzureCrlService` is only responsible for retrieving the initial azure resource. 
Since `ApplicationService` only contained two pieces of functionality, both of which were used exclusively by `AzureService`, that functionality was moved into `AzureService`.

It's possible that functionality over top of the raw resources such as `AzureService.getTenantForSubscription()` and `AzureService. getApplicationsForSubscription()` should be either in a separate abstraction layer, or that the `AzureCrlService` should provide this itself. 
I wanted to get some opinions on this though, and see if anyone had any other ideas.

[WOR-937]: https://broadworkbench.atlassian.net/browse/WOR-937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlight.java
```diff
@@ -48,7 +46,7 @@ public CreateProfileFlight(FlightMap inputParameters, Object applicationContext)
 
     if (CloudPlatform.AZURE == profile.cloudPlatform()) {
       // we can link the profile to the MRG only after the Sam resource has been created
-      addStep(new LinkBillingProfileIdToMrgStep(appService, samService, profile, user));
```
blakery, line 51: `LinkBillingProfileIdToMrgStep` no longer uses any functionality that was contained in `ApplicationService` (even before the changes in this PR), so there's not any actual dependency on `ApplicationService` at all, in this flight.

service/src/main/java/bio/terra/profile/app/configuration/BeanConfig.java
```diff
@@ -14,6 +15,13 @@
 
 @Configuration
 public class BeanConfig {
+  private static final String CLIENT_NAME = "profile";
+
+  @Bean("crlClientConfig")
+  public ClientConfig crlClientConfig() {
```
blakery, line 21: This was the sole shared resource between the Azure parts of `CrlService` and the GCP part. 
Other constructor parameters of the services are set up as beans, so it made sense to do the same here.

service/src/main/java/bio/terra/profile/service/crl/AzureCrlService.java
```diff
@@ -0,0 +1,58 @@
...
+import com.azure.core.management.profile.AzureProfile;
+import com.azure.resourcemanager.costmanagement.CostManagementManager;
+import com.azure.resourcemanager.managedapplications.ApplicationManager;
+import com.azure.resourcemanager.resources.ResourceManager;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AzureCrlService {
```
cahrens, line 15: Does "CRL" stand for "Cloud Resources Library"?

I think I finally worked this out, but it would be helpful to just add some documentation for other folks who are acronym-challenged.

service/src/main/java/bio/terra/profile/service/crl/AzureCrlService.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public ResourceManager getResourceManager(UUID subscriptionId) {
+    return getResourceManager(
+        UUID.fromString(azureConfiguration.managedAppTenantId()), subscriptionId);
+  }
+
+  /** Returns an Azure {@link ResourceManager} configured for use with CRL. */
+  public ResourceManager getResourceManager(UUID tenantId, UUID subscriptionId) {
+    AzureProfile azureProfile =
+        new AzureProfile(tenantId.toString(), subscriptionId.toString(), AzureEnvironment.AZURE);
```
cahrens, line 34: I assume this is existing code, but it's odd that the you have to specify `AzureEnvironment.AZURE` when creating an `AzureProfile`.

service/src/main/java/bio/terra/profile/service/spendreporting/azure/AzureCostManagementQuery.java
```diff
@@ -14,25 +14,21 @@
...
-  private final CrlService crlService;
+  private final AzureCrlService crlService;
 
-  public AzureCostManagementQuery(CrlService crlService) {
+  @Autowired
+  public AzureCostManagementQuery(AzureCrlService crlService) {
     this.crlService = crlService;
   }
 
-  public Response<QueryResult> resourceGroupCostQueryWithResourceIdGrouping(
```
cahrens, line 30: Unused code?

service/src/main/java/bio/terra/profile/service/crl/GcpCrlService.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.stereotype.Component;
+
+/** Provides GCP Cloud resources for the application */
+@Component
+public class GcpCrlService {
+
+  private final ClientConfig clientConfig;
+
+  @Autowired
+  public GcpCrlService(ClientConfig clientConfig) {
```
cahrens, line 20: Can you point me to code that instantiates this? I'm having trouble understanding where that is.

service/src/main/java/bio/terra/profile/service/crl/AzureCrlService.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class AzureCrlService {
+
+  private final AzureConfiguration azureConfiguration;
+  private final ClientConfig clientConfig;
+
+  @Autowired
+  public AzureCrlService(AzureConfiguration azureConfiguration, ClientConfig clientConfig) {
```
cahrens, line 21: Likewise, how does this ultimately get instantiated?

cahrens:
@blakery this LGTM and I really appreciate the excellent PR description. I'm just having a bit of trouble understanding where the instantiation of the CRL instances happens (as the input args are now different and I would expect to see that get updated).

service/src/main/java/bio/terra/profile/service/crl/AzureCrlService.java
blakery: Yeah, I thought so too.
Looking at `AzureEnvironment`, it appears there are a bunch of alternatives specific azure instances, like `AZURE_US_GOVERNMENT`, `AZURE_GERMANY`, and `AZURE_CHINA`.

service/src/main/java/bio/terra/profile/service/spendreporting/azure/AzureCostManagementQuery.java
blakery: yup

service/src/main/java/bio/terra/profile/service/crl/GcpCrlService.java
blakery: It gets created automatically by spring.
This is possible because it's annotated with `@Component`, it has a constructor annotated with `@Autowired`, and there are beans set up for all of the constructor parameters.

cahrens: Ah right, magic!

service/src/main/java/bio/terra/profile/app/configuration/BeanConfig.java
blakery: I moved this piece from `CrlService`; the only change I made was to make it a bean.

The difference is that if we needed to use Janitor, we'd need to add the `CleanupConfig`.

cahrens: Could you update the comment to clarify that the thing it doesn't need is `CleanupConfig`?

blakery:
> @blakery this LGTM and I really appreciate the excellent PR description. I'm just having a bit of trouble understanding where the instantiation of the CRL instances happens (as the input args are now different and I would expect to see that get updated).

It doesn't get _directly_ instantiated anywhere.
It's mostly provided by spring. The `@Component` annotation is kind of like a class level usage of `@Bean`. Whenever you ask spring for an instance of the class, spring will do the instantiation, and populate the constructor params from the configured beans. This is why the method to get the config needed the `@Bean` annotation.

PR #227, opened 2023-06-02
WOR-939 BPM provider tests

blakery:
[WOR-939](https://broadworkbench.atlassian.net/browse/WOR-939)

Set up provider side tests to run against consumer-generated pacts:
* add gradle task to verify pacts, with 
* sets up a reasonable spring test config with appropriate mocks
* pact broker integration with can-i-deploy

[WOR-939]: https://broadworkbench.atlassian.net/browse/WOR-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/verify_consumer_pacts.yml
+++ b/.github/workflows/verify_consumer_pacts.yml
@@ -116,7 +116,7 @@ jobs:
           echo "CURRENT_BRANCH=${GITHUB_REF/refs\/heads\//""}" >> $GITHUB_ENV
           echo "CURRENT_SHA=$GITHUB_SHA" >> $GITHUB_ENV
 
-      - name: "This step will only run when this workflow is triggered by a Pact Broker webhook event"
+      - name: Set environment variables when triggered by a Pact Broker webhook event
         if: ${{ inputs.pb-event-type != '' }}
         run: |
           echo "pb-event-type=${{ inputs.pb-event-type }}"
--- a/.github/workflows/verify_consumer_pacts.yml
+++ b/.github/workflows/verify_consumer_pacts.yml
@@ -12,15 +12,14 @@ name: Verify consumer pacts
 # - PACT_BROKER_PASSWORD - the Pact Broker password
 on:
   pull_request:
-    branches:
-      - develop
-    paths-ignore:
-      - 'README.md'
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
   push:
-    branches:
-      - develop
-    paths-ignore:
-      - 'README.md'
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+  merge_group:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
   workflow_dispatch:
     inputs:
       pb-event-type:
--- a/.github/workflows/verify_consumer_pacts.yml
+++ b/.github/workflows/verify_consumer_pacts.yml
@@ -149,6 +149,21 @@ jobs:
           echo "git rev-parse HEAD"
           git rev-parse HEAD
 
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Gradle cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ github.ref }}-${{ github.sha }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ github.ref }}
+
       - name: Verify consumer pacts and publish verification status to Pact Broker
         id: verification-test
         env:
--- a/.github/workflows/verify_consumer_pacts.yml
+++ b/.github/workflows/verify_consumer_pacts.yml
@@ -67,7 +67,22 @@ env:
   PACT_BROKER_URL: https://pact-broker.dsp-eng-tools.broadinstitute.org
 
 jobs:
+
+  bump-check:
+    runs-on: ubuntu-latest
+    outputs:
+      is-bump: ${{ steps.skiptest.outputs.is-bump }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Skip version bump merges
+        id: skiptest
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
+
   verify-consumer-pact:
+    needs: [ bump-check ]
+    if: needs.bump-check.outputs.is-bump == 'no'
     runs-on: ubuntu-latest
     permissions:
       contents: 'read'
@@ -187,8 +202,8 @@ jobs:
 
   can-i-deploy: # The can-i-deploy job will run as a result of a PR. It reports the pact verification statuses on all deployed environments.
     runs-on: ubuntu-latest
-    if: ${{ inputs.pb-event-type == '' }}
-    needs: [ verify-consumer-pact ]
+    if: ${{ inputs.pb-event-type == '' }} && needs.bump-check.outputs.is-bump == 'no'
+    needs: [ verify-consumer-pact, bump-check ]
     steps:
       - name: Dispatch to terra-github-workflows
         uses: broadinstitute/workflow-dispatch@v3
--- a/.github/workflows/verify_consumer_pacts.yml
+++ b/.github/workflows/verify_consumer_pacts.yml
@@ -1,0 +1,186 @@
+name: Verify consumer pacts
+# The purpose of this workflow is to verify ANY consumer contract(s) dependent on BPM provider using Pact framework.
+#
+# This workflow is triggered when
+#
+# 1. Consumer makes a change that results in a new pact published to Pact Broker (will verify ONLY the changed pact and publish the verification results back to the broker)
+# 2. Provider makes a change (runs verification tests against ALL DEPLOYED consumer pact versions and publishes corresponding verification results)
+#
+#
+# The workflow requires Pact broker credentials
+# - PACT_BROKER_USERNAME - the Pact Broker username
+# - PACT_BROKER_PASSWORD - the Pact Broker password
+on:
+  pull_request:
+    branches:
+      - develop
+    paths-ignore:
+      - 'README.md'
+  push:
+    branches:
+      - develop
+    paths-ignore:
+      - 'README.md'
+  workflow_dispatch:
+    inputs:
+      pb-event-type:
+        description: 'the Pact Broker event type that triggers this workflow'
+        required: true
+        type: string
+      consumer-name:
+        description: 'the consumer name'
+        required: true
+        type: string
+      consumer-version-number:
+        description: 'the version number of the most recent consumer version associated with the pact content'
+        required: true
+        type: string
+      provider-version-number:
+        description: 'the provider version number for the verification result'
+        required: false
+        type: string
+      consumer-version-tags:
+        description: 'the list of tag names for the most recent consumer version associated with the pact content, separated by ", "'
+        required: true
+        type: string
+      consumer-version-branch:
+        description: 'the name of the branch for most recent consumer version associated with the pact content'
+        required: true
+        type: string
+      provider-version-branch:
+        description: 'the name of the branch for the provider version associated with the verification result'
+        required: false
+        type: string
+      consumer-labels:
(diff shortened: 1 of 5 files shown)


service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -0,0 +1,137 @@
...
+      @ComponentScan.Filter(
+          type = FilterType.REGEX,
+          pattern = {"jdbcDialect", "jdbcCustomConversions", "jdbcMappingContext"})
+    })
+@ConfigurationPropertiesScan(basePackages = {"bio.terra.profile"})
+public class BPMProviderTest {
+
+  @LocalServerPort int port;
+
+  @MockBean AzureCrlService azureCrlService;
```
blakery, line 74: At the current level of tests, most of mocks are just so spring has something to instantiate beans with when creating a context.
As we add tests, if we wanted to push the mocks lower down, we can: we just need to keep in mind that these tests are about testing the api-level interaction between BPM and consuming services, not about testing the internal logic of BPM.

service/build.gradle
```diff
@@ -39,7 +39,7 @@ dependencies {
 	// Cloud Resource Library
 	implementation group: "com.azure.resourcemanager", name: "azure-resourcemanager-managedapplications", version: "1.0.0-beta.2"
 	implementation group: "com.azure.resourcemanager", name: "azure-resourcemanager-subscription", version: "1.0.0-beta.2"
-	implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-costmanagement', version: "1.0.0-beta.6"
+	implementation group: 'com.azure.resourcemanager', name: 'azure-resourcemanager-costmanagement', version: "1.0.0-beta.5"
```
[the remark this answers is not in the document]
blakery: No: that's a mistake
There were some conflicts when I rebased, with the dependency updates.
I thought I got all of them, but I missed that one.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @TestTemplate
+  @ExtendWith(PactVerificationSpringProvider.class)
+  void verifyPact(PactVerificationContext context) {
+    context.verifyInteraction();
+  }
+
+  @State("a GCP billing profile")
+  Map<String, Object> gcpBillingProfileState() throws InterruptedException {
```
marctalbott, line 100: Is it possible to have both Azure and GCP billing profiles in BPM's state? From what I can tell, it's either one or the other at the moment

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -0,0 +1,137 @@
...
+    basePackages = {"bio.terra.profile.app.controller", "bio.terra.profile.service"},
+    excludeFilters = {
+      @ComponentScan.Filter(
+          type = FilterType.ASSIGNABLE_TYPE,
+          value = {BeanConfig.class, ProfileDao.class}),
+      @ComponentScan.Filter(
+          type = FilterType.REGEX,
+          pattern = {"jdbcDialect", "jdbcCustomConversions", "jdbcMappingContext"})
+    })
+@ConfigurationPropertiesScan(basePackages = {"bio.terra.profile"})
```
marctalbott, line 69: There's a lot of annotations here and seemingly a lot more than we need for other tests, could you give a little information about what these all do and why they're needed?

service/src/test/java/bio/terra/profile/pact/provider/ProviderStateData.java
```diff
@@ -0,0 +1,42 @@
...
+          UUID.fromString("479e6559-027c-43b3-b422-09c18dcca3c8"),
+          "GCP Profile",
+          "a GCP billing profile",
+          "a biller?",
+          CloudPlatform.GCP,
+          Optional.of("some-billing-account"),
+          Optional.empty(),
+          Optional.empty(),
+          Optional.empty(),
+          Instant.now(),
```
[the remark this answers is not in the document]
marctalbott: I'm not fully grokking this, do we need to hardcode values or no?

marctalbott (commented):
Generally looks okay, but I have a few questions as I'm not sure I'm fully understanding all of this

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.springframework.cache.CacheManager;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.FilterType;
+
+@Tag("provider-test")
+@Provider("bpm-provider")
+@PactBroker
+// for local testing, put any test pacts in the service/pacts folder.
+// then comment out the above line, and uncomment the following line
+// @PactFolder("pacts")
```
aherbst-broad, line 49: Huh, it's annoying pact can let you control this via an env var or some other config.

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,200 @@
...
+#
+# The workflow requires Pact broker credentials
+# - PACT_BROKER_USERNAME - the Pact Broker username
+# - PACT_BROKER_PASSWORD - the Pact Broker password
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore: [ '**.md' ]
+  push:
+    branches: [ main ]
```
aherbst-broad, line 18: Does this mean it'll run on all merges to main including version bumps? It might be good to include a bump check similar to what we do in a few our other actions.

aherbst-broad (approved):

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
blakery: At present, that is true.
It _is_ possible to specify multiple states in a single pact (and it's also useful), so that is something I think we need to figure out.

Since the focus on this was the infrastructure integration, and working spring setup, I figured that this was adequate for the current tests to pass, and I didn't want to get bogged down in the details of figuring out mockito setup, and exactly which pact hook is the best place to do that.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
blakery: There are some of these that, going back though and testing them out, are no longer necessary, because of some iteration on the spring configuration. That would be the `@ComponentScan` and `@ConfigurationPropertiesScan`. I'll remove those.

`@EnableAutoConfiguration` is actually already present implicitly on other tests, it's just coming by way in inheriting from `BaseSpringUnitTest`, which is getting its configuration from `Main`. This is just adding an additional auto-config class to ignore.

The remainder are for pact

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
blakery: there _may_ be a way, but I haven't found it yet

service/src/test/java/bio/terra/profile/pact/provider/ProviderStateData.java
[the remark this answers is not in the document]
blakery: It _does_ matter on the consumer side, but I don't think it matters on the provider side.

.github/workflows/verify_consumer_pacts.yml
cahrens: Looking below, it seems that the workflow is a no-op for version bumps. https://github.com/DataBiosphere/terra-billing-profile-manager/pull/227/files#diff-88f8d59210f317b7bdd12e6a807c0b84cf473372221acb3ce860f5a0160183b1R85

.github/workflows/verify_consumer_pacts.yml
```diff
@@ -0,0 +1,215 @@
...
+          echo "env.CURRENT_BRANCH=${{ env.CURRENT_BRANCH }}   # This reflects the branch being checked out if CHECKOUT_BRANCH is empty"
+          echo "env.CURRENT_SHA=${{ env.CURRENT_SHA }}         # This reflects the git commit hash of the branch being checked out if CHECKOUT_BRANCH is empty"
+          echo "env.PROVIDER_BRANCH=${{ env.PROVIDER_BRANCH }} # This reflects the provider branch for pact verification"
+          echo "env.PROVIDER_SHA=${{ env.PROVIDER_SHA }}       # This reflects the provider version for pact verification"
+          echo "env.CONSUMER_BRANCH=${{ env.CONSUMER_BRANCH }} # This reflects the consumer branch for pact verification (generated by Pact Broker)"
+          echo "env.CONSUMER_SHA=${{ env.CONSUMER_SHA }}       # This reflects the consumer version for pact verification (generated by Pact Broker)"
+          ./gradlew --build-cache verifyPacts
+
+
+  can-i-deploy: # The can-i-deploy job will run as a result of a PR. It reports the pact verification statuses on all deployed environments.
```
cahrens, line 203: I'm not clear what this means. If there is a failure, does it prevent deployment?

service/build.gradle
```diff
@@ -73,3 +76,18 @@ tasks.register("pactTests", Test) {
 		includeTags "pact-test"
 	}
 }
+
+tasks.register("verifyPacts", Test) {
+	useJUnitPlatform {
```
cahrens, line 81: Out of curiosity, why the crazy indenting?

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
cahrens: It would be worth putting a comment in the test that the mocks are only there so that spring can instantiate beans when creating a context.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -0,0 +1,122 @@
...
+    context.verifyInteraction();
+  }
+
+  @State("a GCP billing profile")
+  Map<String, Object> gcpBillingProfileState() throws InterruptedException {
+    var profile = ProviderStateData.gcpBillingProfile;
+    setUpProfileDaoGets(List.of(profile));
+    when(samService.hasActions(any(), eq(SamResourceType.PROFILE), eq(profile.id())))
+        .thenReturn(true);
+    return Map.of(
```
cahrens, line 90: To see if I'm understanding… this is saying that when a consumer asks for a GCP Billing profile, the only things that it uses in the response are the `gcpProfileId` and the `billingAccountId`?

.github/workflows/verify_consumer_pacts.yml
blakery: I added that based on this conversation.

service/build.gradle
blakery: It's the same as the rest of the file
Although I _did_ just realizes that this file uses tabs instead of spaces. eww....

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
blakery: No; It just means that these are specific values that are available in the consumer pact.

For instance, if you are testing retrieving the profile by id (which in fact is what the WSM tests are doing), you want to ensure that the returned object has the same id as what you requested.
But you need to know what id to query. So if you specify `"${gcpProfileId}"` in the consumer test, this is what you'll get. Then you know both what id to use in the query, and what use for comparisons.

.github/workflows/verify_consumer_pacts.yml
blakery: It means a red X on the PR (so it blocks merging).

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
cahrens: But isn't something supposed to verify that the provider returns the parameters that the code depends on?

cahrens (approved):
Blake explained my question "in person"

PR #236, opened 2023-06-23
Add provider contract for status check (WOR-908).

cahrens:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-908

Provider-side pact test to be used by https://github.com/broadinstitute/rawls/pull/2400.

Note that more interesting verifications will follow… for my first I chose the least complicated endpoint.

### DIFF
--- a/service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
+++ b/service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
@@ -1,6 +1,7 @@
 package bio.terra.profile.pact.provider;
 
 import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.when;
 
 import au.com.dius.pact.provider.junit5.HttpTestTarget;
@@ -37,6 +38,9 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.boot.test.web.server.LocalServerPort;
 import org.springframework.cache.CacheManager;
+import org.springframework.jdbc.core.ConnectionCallback;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 @Tag("provider-test")
 @Provider("bpm-provider")
@@ -68,6 +72,10 @@ public class BPMProviderTest {
   @MockBean GcpCrlService gcpCrlService;
   @MockBean JobService jobService;
   @MockBean CacheManager cacheManager;
+
+  // jdbcTemplate beans are used for profileStatusService when checking CloudSQL status
+  @MockBean NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+  @MockBean JdbcTemplate jdbcTemplate;
   @Autowired ProfileStatusService profileStatusService;
 
   @BeforeEach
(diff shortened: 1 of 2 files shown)


service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -29,20 +30,24 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 @Tag("provider-test")
 @Provider("bpm-provider")
 @PactBroker
 // for local testing, put any test pacts in the service/pacts folder.
 // then comment out the above line, and uncomment the following line
-// @PactFolder("pacts")
+// @PactFolder("src/test/java/bio/terra/profile/service/pacts")
```
cahrens, line 50: The previous path was incorrect.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -83,6 +90,16 @@ void verifyPact(PactVerificationContext context) {
...
+  @State("BPM is ok")
+  void getBpmStatus() {
+    when(samService.status()).thenReturn(new SystemStatusSystems().ok(true));
+    // For CloudSQL return value
+    doReturn(jdbcTemplate).when(namedParameterJdbcTemplate).getJdbcTemplate();
+    doReturn(true).when(jdbcTemplate).execute(any(ConnectionCallback.class));
+    // Checks status of services and stores result for the next time status endpoint is called
+    profileStatusService.checkStatus();
+  }
+
```
cahrens, line 102: FYI, this is what the pact file looks like that is generated by the Rawls consumer test:

```
{
  "consumer": {
    "name": "rawls-consumer"
  },
  "interactions": [
    {
      "description": "Request to BPM ok status",
      "providerStates": [
        {
          "name": "BPM is ok",
          "params": {
            "CloudSQL": "ok",
            "Sam": "ok"
          }
        }
      ],
      "request": {
        "headers": {
          "Accept": "application/json"
        },
        "method": "GET",
        "path": "/status"
      },
      "response": {
        "body": {
          "ok": true,
          "systems": {
            "CloudSQL": {
              "ok": true
            },
            "Sam": {
              "ok": true
            }
          }
        },
        "headers": {
          "Content-type": "application/json"
        },
        "matchingRules": {
          "body": {
            "$.ok": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "type"
                }
              ]
            },
            "$.systems.CloudSQL.ok": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "type"
                }
              ]
            },
            "$.systems.Sam.ok": {
              "combine": "AND",
              "matchers": [
                {
                  "match": "type"
                }
              ]
            }
          }
        },
        "status": 200
      }
    }
  ],
  "metadata": {
    "pact-jvm": {
      "version": "4.4.0"
    },
    "pactSpecification": {
      "version": "3.0.0"
    }
  },
  "provider": {
    "name": "bpm-provider"
  }
}

File: service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java [hunk 3]
```diff
@@ -83,6 +90,16 @@ void verifyPact(PactVerificationContext context) {
     context.verifyInteraction();
   }
 
+  @State("BPM is ok")
+  void getBpmStatus() {
+    when(samService.status()).thenReturn(new SystemStatusSystems().ok(true));
+    // For CloudSQL return value
+    doReturn(jdbcTemplate).when(namedParameterJdbcTemplate).getJdbcTemplate();
```
blakery, line 97: Now I see why I could never get that to work - I didn't realize I needed to proxy that method.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
blakery: I didn't really intend for it to be in the source code (since it isn't a java file).
The original path will work for something in the root directory of the service project.

My suggestion is to use the `src/test/resources` folder, as something that would both make sense in terms of normal project organizational conventions, and be more obvious as to the location being referenced.

blakery (approved):

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
cahrens: OK, when I loop back around to this file for the next set of endpoints I will move this elsewhere.

sergiygetlin (approved):

PR #240, opened 2023-06-29
Additional test cases (WOR-908).

cahrens:
This implements the remainder of the provider support for https://github.com/broadinstitute/rawls/pull/2420.

Ticket: https://broadworkbench.atlassian.net/browse/WOR-908

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -53,9 +53,9 @@ dependencies {
 	// Allows us to mock final classes
 	testImplementation 'org.mockito:mockito-inline:5.2.0'
 
-	testImplementation 'au.com.dius.pact.consumer:junit5:4.1.7'
-	testImplementation("au.com.dius.pact.provider:junit5:4.1.7")
-	testImplementation("au.com.dius.pact.provider:junit5spring:4.1.7")
+	testImplementation 'au.com.dius.pact.consumer:junit5:4.6.1'
+	testImplementation("au.com.dius.pact.provider:junit5:4.6.1")
+	testImplementation("au.com.dius.pact.provider:junit5spring:4.6.1")
 }
 test {
 	systemProperties['pact.rootDir'] = "$buildDir/pacts"
--- a/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
@@ -82,7 +82,7 @@ public class ProfileApiController implements ProfileApi {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     SamPolicyModel policy =
         profileService.addProfilePolicyMember(id, policyName, requestBody.getEmail(), user);
-    return new ResponseEntity<>(policy, HttpStatus.OK);
+    return new ResponseEntity<>(policy, HttpStatus.CREATED);
   }
 
   @Override
--- a/service/src/test/java/bio/terra/profile/pact/SamServiceTest.java
+++ b/service/src/test/java/bio/terra/profile/pact/SamServiceTest.java
@@ -5,21 +5,19 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import au.com.dius.pact.consumer.MockServer;
 import au.com.dius.pact.consumer.dsl.PactDslJsonBody;
 import au.com.dius.pact.consumer.dsl.PactDslWithProvider;
-import au.com.dius.pact.consumer.junit5.PactConsumerTestExt;
+import au.com.dius.pact.consumer.junit5.PactConsumerTest;
 import au.com.dius.pact.consumer.junit5.PactTestFor;
 import au.com.dius.pact.core.model.PactSpecVersion;
 import au.com.dius.pact.core.model.RequestResponsePact;
 import au.com.dius.pact.core.model.annotations.Pact;
 import bio.terra.profile.app.configuration.SamConfiguration;
-import bio.terra.profile.common.BaseUnitTest;
 import bio.terra.profile.service.iam.SamService;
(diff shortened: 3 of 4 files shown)


service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
```diff
@@ -82,7 +82,7 @@ public ResponseEntity<SamPolicyModel> addProfilePolicyMember(
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     SamPolicyModel policy =
         profileService.addProfilePolicyMember(id, policyName, requestBody.getEmail(), user);
-    return new ResponseEntity<>(policy, HttpStatus.OK);
+    return new ResponseEntity<>(policy, HttpStatus.CREATED);
```
cahrens, line 85: While writing the test, I noticed that Swagger advertised 201, but we were returning 200. I think 201 is more correct, so changing this to match Swagger.

I looked through the Rawls usages of this method and they don't check the status code. I am hoping this is a low risk change.

service/src/test/java/bio/terra/profile/pact/SamServiceTest.java
```diff
@@ -5,20 +5,19 @@
...
 import bio.terra.profile.service.iam.SamService;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
 
 @Tag("pact-test")
-@ExtendWith(PactConsumerTestExt.class)
-public class SamServiceTest extends BaseUnitTest {
+@PactConsumerTest
+public class SamServiceTest {
```
cahrens, line 20: Because this extended `BaseUnitTest`, the test was being executed during the normal unit test phase (as well as during the pact test phase). With newer versions of pact, running it as a regular unit test failed.

The example pact tests do not extend anything, and I verified that the cases still run during the pact test phase.

service/src/test/java/bio/terra/profile/pact/SamServiceTest.java
```diff
@@ -51,7 +50,7 @@ public RequestResponsePact userStatusPact(PactDslWithProvider builder) {
   }
 
   @Test
-  @PactTestFor(pactMethod = "statusApiPact")
+  @PactTestFor(pactMethod = "statusApiPact", pactVersion = PactSpecVersion.V3)
```
cahrens, line 53: Specifying `pactVersion` fixed the failures we were seeing on the dependabot PR.

service/build.gradle
```diff
@@ -53,9 +53,9 @@ dependencies {
 	// Allows us to mock final classes
 	testImplementation 'org.mockito:mockito-inline:5.2.0'
 
-	testImplementation 'au.com.dius.pact.consumer:junit5:4.1.7'
-	testImplementation("au.com.dius.pact.provider:junit5:4.1.7")
-	testImplementation("au.com.dius.pact.provider:junit5spring:4.1.7")
+	testImplementation 'au.com.dius.pact.consumer:junit5:4.6.1'
```
cahrens, line 56: The point release dependabot PR was failing. I decided to go to the newest version while fixing the compatibility issues.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -124,6 +142,111 @@ Map<String, Object> azureBillingProfileState() throws InterruptedException {
...
+    return Map.of("subscriptionId", subscriptionId.toString());
+  }
+
+  @State("a JobService that supports profile creation")
+  Map<String, Object> creationJobServiceExistsState() {
+    var jobBuilder = mock(JobBuilder.class);
+
+    var profile = ProviderStateData.azureBillingProfile;
+
+    when(jobService.newJob()).thenReturn(jobBuilder);
```
cahrens, line 179: I copied this mocking from the unittest for profile creation. I first tried to create a more "real" JobService so that more of the actual flight code would execute, but that proved to be impossible with all the of Stairway dependencies.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
```diff
@@ -124,6 +142,111 @@ Map<String, Object> azureBillingProfileState() throws InterruptedException {
...
+  }
+
+  @State("a JobService that supports profile deletion")
+  void deletionJobServiceExistsState() {
+    var jobBuilder = mock(JobBuilder.class);
+    String jobId = "jobId";
+
+    var profile = ProviderStateData.azureBillingProfile;
+
+    when(jobService.newJob()).thenReturn(jobBuilder);
```
cahrens, line 199: Mocking copied from profile deletion unittest.

service/src/test/java/bio/terra/profile/pact/SamServiceTest.java
blakery: weird... I don't think that was in any of the docs, tutorials, stackexchange threads, or arcane tombs I read about pact.

service/src/test/java/bio/terra/profile/pact/provider/BPMProviderTest.java
blakery: I went down the same path, with the same result

blakery (approved):
Really glad you sorted out updating the pact version

service/src/test/java/bio/terra/profile/pact/SamServiceTest.java
cahrens: It's in the examples now….

aherbst-broad (approved):

PR #273, opened 2023-08-28
WOR-548: BPM Smoke Tests

blakery:
Checks that calls succeed to the following endpoints:
* `GET /status` - the service is available, and functional according to the service itself 
* `GET /api/profiles/v1` - basic database connection is functional
* `GET /api/azure/v1/managedApps` - basic Azure connection is functional


smoke_tests/tests/authenticated/azure/managed_apps_tests.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class ManagedAppsTests(SmokeTestCase):
+    AZURE_SUBSCRIPTION_ID = None
+
+    @staticmethod
+    def managed_apps_url() -> str:
+        return SmokeTestCase.build_bpm_url("/api/azure/v1/managedApps")
+
+    def test_retrieving_managed_apps(self):
```
blakery, line 13: Even if there are no available managed apps, this should fail if the service can't query Azure.
Any functionality such as it returning the correct apps for the user, filtering out unavailable apps unless specified, etc, should all be tested by other means.

smoke_tests/tests/authenticated/azure/managed_apps_tests.py
```diff
@@ -0,0 +1,25 @@
...
+        params = {"azureSubscriptionId": self.AZURE_SUBSCRIPTION_ID}
+        response = SmokeTestCase.call_bpm(self.managed_apps_url(), frozenset(params.items()),
+                                          SmokeTestCase.USER_TOKEN)
+        self.assertEqual(response.status_code, 200)
+        body = json.loads(response.text)
+        apps = body["managedApps"]
+        # Note that this doesn't check that there are any number of items in the list of managed apps
+        # only that any there have this form
+        for app in apps:
```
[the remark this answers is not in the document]
aherbst-broad: Probably fine. Wonder if this could be a more basic existence check like `assertTrue("subscriptionId" in app)`.

smoke_tests/tests/smoke_test_case.py
```diff
@@ -0,0 +1,27 @@
+import re
+from functools import cache
+from unittest import TestCase
+from urllib.parse import urljoin
+
+import requests
+from requests import Response
+
+
+class SmokeTestCase(TestCase):
```
aherbst-broad, line 10: Not a huge deal but I generally prefer pytest's decorator-based approach to unittesting vs. the class-based approach w/unittest. Something for tech talk another time.

smoke_tests/tests/smoke_test_case.py
```diff
@@ -0,0 +1,27 @@
...
+    @staticmethod
+    def build_bpm_url(path: str) -> str:
+        assert SmokeTestCase.BPM_HOST, "ERROR - BPMSmokeTests.BPM_HOST not properly set"
+        if re.match(r"^\s*https?://", SmokeTestCase.BPM_HOST):
+            return urljoin(SmokeTestCase.BPM_HOST, path)
+        else:
+            return urljoin(f"https://{SmokeTestCase.BPM_HOST}", path)
+
+    @staticmethod
+    @cache
```
aherbst-broad, line 23: Is this caching responses and if so do we want that for a smoke test harness like this?

aherbst-broad (commented):
Looking good, had a few questions. How will we run these?

smoke_tests/tests/smoke_test_case.py
blakery: If I understand it correctly: yes.
The idea is that these should never be doing anything complex, such as expecting different responses for the same input. 
On the other hand, it is convenient for separating out the individual checks we're making on responses for sake of clarity and readability (such as testing the status code and sub-systems separately for status endpoint).

smoke_tests/tests/smoke_test_case.py
```diff
@@ -0,0 +1,27 @@
...
+import requests
+from requests import Response
+
+
+class SmokeTestCase(TestCase):
+    BPM_HOST = None
+    USER_TOKEN = None
+
+    @staticmethod
+    def build_bpm_url(path: str) -> str:
```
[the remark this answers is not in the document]
blakery: I agree. There's almost certainly something available to generate from swagger.

blakery:
> Looking good, had a few questions. How will we run these?

The _purpose_ of these is a low overhead way to check if the service is basically operational in a given environment.
They are intended to be run either as the final step in the deployment check list for releasing the service independently, or ad-hoc if we need it. 

So the immediate intent is for us to run them directly. 

It sounds like there's been some tooling being added (in beehive?) to run things like this (on release?), so we might want to add a github workflow to support that. 
But that will also require getting a user token to make the calls, passing in a subscription id, if we want the managedApps check to run, etc. But I think this supports what we need now, and the automation is a future problem.

smoke_tests/tests/smoke_test_case.py
aherbst-broad: Let me ask a different way: how does this simplify things? IME caching adds performance at the expense of complexity--are there performance concerns?

aherbst-broad:
@blakery thanks, makes sense. Can you enhance the readme or put up a doc somewhere that enumerates the steps we (the workspaces team) will need to go through when running these as part of our SOP?

smoke_tests/smoke_tests.py
```diff
@@ -0,0 +1,92 @@
...
+from tests.authenticated.billing_profiles import BillingProfileTests
+
+DESCRIPTION = """
+BPM Smoke Test
+Enter the host (domain and optional port) of the BPM instance you want to to test. This test will ensure that the BPM 
+instance running on that host is minimally functional.
+"""
+
+
+def gather_tests(is_authenticated: bool = False, azure_subscription: bool = False) -> TestSuite:
```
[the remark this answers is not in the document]
cahrens: And actually you are passing the auth token for is_authenticated, aren't you?

Probably fine to leave them as booleans, but make the actual values passed to them booleans.

smoke_tests/smoke_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+def main(main_args):
+    if main_args.user_token:
+        verify_user_token(main_args.user_token)
+
+    SmokeTestCase.BPM_HOST = main_args.bpm_host
+    SmokeTestCase.USER_TOKEN = main_args.user_token
+    ManagedAppsTests.AZURE_SUBSCRIPTION_ID = main_args.azure_sub_id
+
+    test_suite = gather_tests(main_args.user_token, main_args.azure_sub_id)
```
cahrens, line 40: As mentioned above, this would make more sense to me if you made the arguments passed booleans, like suggested below. Or perhaps return a boolean from `verify_user_token` to represent that the use has authenticated and use that for the first argument?

```
test_suite = gather_tests(main_args.user_token != null, main_args.azure_sub_id != null)

File: smoke_tests/smoke_tests.py [hunk 8]
```diff
@@ -0,0 +1,92 @@
...
+    parser.add_argument(
+        "user_token",
+        nargs='?',
+        default=None,
+        help="Optional. If present, will test additional authenticated endpoints using the specified token"
+    )
+    parser.add_argument(
+        "--azure_sub_id",
+        type=str,
+        help="Optional Azure SubscriptionId. If present, will test retrieving azure managed apps using the specified id"
```
cahrens, line 69: default value since it is optional? For `user_token` you specified a default of None.

smoke_tests/smoke_tests.py
```diff
@@ -0,0 +1,92 @@
...
+        prog="BPM Smoke Tests",
+        description=DESCRIPTION,
+    )
+    parser.add_argument(
+        "bpm_host",
+        help="Required domain with optional port number of the BPM host you want to test"
+    )
+    parser.add_argument(
+        "user_token",
+        nargs='?',
```
cahrens, line 62: What is `nargs='?'` for? And should we say `type=str`?

smoke_tests/smoke_tests.py
cahrens: Why is the name `--azure_sub_id`, whereas `user_token` didn't have the `--` prefix?

smoke_tests/tests/authenticated/azure/managed_apps_tests.py
```diff
@@ -0,0 +1,25 @@
...
+
+class ManagedAppsTests(SmokeTestCase):
+    AZURE_SUBSCRIPTION_ID = None
+
+    @staticmethod
+    def managed_apps_url() -> str:
+        return SmokeTestCase.build_bpm_url("/api/azure/v1/managedApps")
+
+    def test_retrieving_managed_apps(self):
+        params = {"azureSubscriptionId": self.AZURE_SUBSCRIPTION_ID}
```
cahrens, line 14: Should we assert first that `self.AZURE_SUBSCRIPTION_ID` is not None?

cahrens: And possibly `USER_TOKEN`?

smoke_tests/smoke_tests.py
blakery: `user_token` is positional, where `--azure_sub_id` is not. 
I wanted to limit the number of of positional arguments generally, since too many of them, and more than one positional argument can get confusing quickly.
`user_token` is related to a much broader class of functionality: without it you can't test _any_ authenticated API, where azure_sub_id is a more specific functionality, so I felt it made sense for that to be positional/

cahrens: That makes sense, thanks.

smoke_tests/smoke_tests.py
blakery: It means this is a positional argument, and that the argparser will consume one argument from the command line, if possible and produce a single item. If no command-line argument is present, it will produce the default value (which is why we're specifying the default of `None`).

Adding `type=str` is a good idea, to keep things consistent.

smoke_tests/tests/authenticated/azure/managed_apps_tests.py
blakery: I figured we were doing that when setting up the test case.

smoke_tests/smoke_tests.py
blakery: This is also why `--azure_sub_id` doesn't have default value - only positional arguments need to specify them.

smoke_tests/smoke_tests.py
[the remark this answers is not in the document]
blakery: I was just relying on python's type coercion. If it's there it will be truthy, if it's not, it will be falsey.

 I don't mind cleaning it up to be more type-consistent.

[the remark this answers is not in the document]
cahrens: If you are going to specify the types as booleans, then make them booleans. If you just want to leave off the types then I'm also OK with that… it just gets confusing to me when you say it's a type that it is not.

smoke_tests/README.md
```diff
@@ -0,0 +1,30 @@
+
+
+## Purpose
+To provide a low overhead way to check that BPM is basically operational in a given environment.
```
aherbst-broad, line 4: Nit: `s/basically//g` It's either operational or it isn't.

smoke_tests/smoke_tests.py
```diff
@@ -0,0 +1,94 @@
...
+            suite.addTests(managed_apps_tests)
+    return suite
+
+
+def main(main_args):
+    SmokeTestCase.BPM_HOST = main_args.bpm_host
+    SmokeTestCase.USER_TOKEN = main_args.user_token
+    ManagedAppsTests.AZURE_SUBSCRIPTION_ID = main_args.azure_sub_id
+
+    valid_user_token = main_args.user_token is not None and verify_user_token(main_args.user_token)
```
aherbst-broad, line 37: FWIW this can be done programmatically https://github.com/broadinstitute/dsp-scripts/blob/master/azure/workspaces/tools/utils/auth.py#L20

aherbst-broad (approved):

smoke_tests/README.md
blakery: I meant that in terms of "the basic functionality is operational", not "the functionality is basically operational".

smoke_tests/tests/smoke_test_case.py
aherbst-broad: I'd say remove it then.

smoke_tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+* Get a shell in the created venv to run the scripts: `poetry shell`
+* Run the test: `python smoke_tests.py <args>`
+* For full usage information: `python smoke_tests.py -h`
+
+
+## Running as part of post-deployment steps:
+* Do setup as described above above
+* Get a user access token (any user that has access to BPM should be sufficient). Eg: `gcloud auth print-access-token`
+* Get an Azure subscription id
+* Run script in poetry shell: `python smoke_tests.py "${BPM_HOST}" "${USER_ACCESS_TOKEN}"  --azure-sub-id="${SUBSCRIPTION_ID}`
```
cahrens, line 31: Should be `--azure_sub_id`.

blakery: right.... the dashes were an initial implementation, but it complicated argparse slightly

PR #279, opened 2023-09-15
[PF-2972] Update dependencies to address security vulnerabilities

zloery:
Verily security has started using a different dependency analysis tool which flagged a handful of vulnerabilities. I'd like to update dependencies to fix issues raised by both Verily's tool and Sourceclear.

I verified the dependency versions are changed by generating lockfiles locally, though it looks like this service generally isn't using them anymore so I didn't include them in this PR.


marctalbott (approved):

rtitle (approved):
LGTM

PR #314, opened 2023-11-07
DDO-3198 Add Metadata Yaml File to  to ingest bpm into DSP backstage

mflinn-broad:
This is a functional no-op PR that just adds a metadata file to the bpm repo that is used to ingest bpm into [DSP's backstage instance](https://backstage.dsp-devops.broadinstitute.org/catalog?filters%5Bkind%5D=component&filters%5Buser%5D=owned). Backstage is an internal platform tool that can be used for a lot of SDLC processes and improving Dev experience. DevOps is just starting to experiment with this as a means to make getting a new service stood up in the terra environments and through the prod readiness checklist easier.

But we also want all our existing services to be in the system so they can get any benefits.

### DIFF
--- a/foundation.yaml
+++ b/foundation.yaml
@@ -20,6 +20,8 @@ spec:
   lifecycle: production
   owner: broadworkspaces
   system: terra
+  dependsOn:
+    - component:sam
   providesApis:
     - terra-billing-profile-manager-api
 ---
--- a/foundation.yaml
+++ b/foundation.yaml
@@ -41,7 +41,7 @@ metadata:
     github.com/project-slug: databiosphere/terra-billing-profile-manager
 spec:
   type: openapi
-  lifecycle: experimental
+  lifecycle: production
   system: terra
   owner: broadworkspaces
   definition:
--- a/foundation.yaml
+++ b/foundation.yaml
@@ -1,0 +1,47 @@
+---
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: terra-billing-profile-manager
+  description: |
+    Terra Billing Profile Manager provides an API 
+    to set up and control access to billing within 
+    Terra across cloud platforms.
+  tags:
+    - java
+    - dsp
+    - terra
+    - springboot
+    - broadworkspaces
+  annotations:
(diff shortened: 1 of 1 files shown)


aherbst-broad (approved):

foundation.yaml
```diff
@@ -0,0 +1,49 @@
...
+  lifecycle: production
+  owner: broadworkspaces
+  system: terra
+  dependsOn:
+    - component:sam
+  providesApis:
+    - terra-billing-profile-manager-api
+---
+apiVersion: backstage.io/v1alpha1
+kind: API
```
cahrens, line 29: What is the difference between the API and the Component kind? Is this supposed to represent a published client vs. the REST API?

mflinn-broad: Sort of. So backstage the tool itself is very flexible since it's a big open source project it can represent lots of different types of software components. The way I would think of it is that `Component` represents BPM as an executable piece of software that runs somewhere. `API` then expresses that the BPM component exposes a REST API that can be consumed by clients. The backstage UI uses this information to do things like render a preview of the swagger page etc. 

The reason for the delineation is that organizations could theoretically have lots of different types of software components other than just backend rest APIs, there could be gRPC services, things that just read off and publish to message queues etc, other things that don't expose APIs at all etc..

foundation.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - java
+    - dsp
+    - terra
+    - springboot
+    - broadworkspaces
+  annotations:
+    github.com/project-slug: databiosphere/terra-billing-profile-manager
+spec:
+  type: openapi
+  lifecycle: experimental
```
cahrens, line 44: Why is the lifecycle experimental in this one (and production []([here](https://github.com/DataBiosphere/terra-workspace-manager/pull/1522/files#diff-4ac81207abfa7187f9f338eaa84e8800dbce73789e22c14713f4997a3849cd95R47)))?

mflinn-broad: copy paste mistake, thanks!

mflinn-broad: fixed

cahrens (approved):

PR #315, opened 2023-11-07
[WOR-1037] Add billing profile policy support

marctalbott:
Ticket: [WOR-1037](https://broadworkbench.atlassian.net/browse/WOR-1037)
* Port over TPS code from WSM
* Support setting policies at profile creation time
* Return policies when getting profiles, delete policies when deleting profiles

[WOR-1037]: https://broadworkbench.atlassian.net/browse/WOR-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/bio/terra/profile/service/profile/model/BillingProfile.java
```diff
@@ -22,7 +24,8 @@ public record BillingProfile(
     Optional<String> managedResourceGroupId,
     Instant createdTime,
     Instant lastModified,
-    String createdBy) {
+    String createdBy,
+    Optional<TpsPolicyInputs> policies) {
```
marctalbott, line 28: I toyed with other options but ended up landing on adding this optional policies value to the `BillingProfile` record. The downside of this is that `BillingProfile` is no longer purely composed of information stored in BPM's DB -- we need to stitch in values from TPS to have a complete representation of the profile. 

Another option would be to introduce another record (e.g. `record BillingProfileDescription(BillingProfile profile, Optional<TpsPolicyInputs> policies)` and returning that from `getPolicies` and `listPolicies` instead. I'm undecided on whether this would be better or worse. One downside is it adds another class that represents a profile in BPM, but it does leave `BillingProfile` untouched. 

Open to input on this!

service/build.gradle
```diff
@@ -50,7 +50,11 @@ dependencies {
 		implementation('org.json:json:20231013')
 	}
 
-	// Test deps
+    // TPS
+    implementation group: "bio.terra", name:"terra-policy-client", version:"1.0.5-SNAPSHOT"
```
aherbst-broad, line 54: Is there a non-snapshot dependency we can pull in?

service/src/main/java/bio/terra/profile/service/profile/model/BillingProfile.java
aherbst-broad: I'd prefer to have a synthetic higher-level record and keep the database models pure. WSM does this with the[ WorkspaceDescription ](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/workspace/model/WorkspaceDescription.java)class, and then transforms _that_ into to the top level API response.

service/src/main/java/bio/terra/profile/service/policy/TpsApiDispatch.java
```diff
@@ -0,0 +1,112 @@
...
+  public void deletePao(UUID objectId) throws InterruptedException {
+    TpsApi tpsApi = policyApi();
+    try {
+      try {
+        TpsRetry.retry(() -> tpsApi.deletePao(objectId));
+      } catch (ApiException e) {
+        throw convertApiException(e);
+      }
+    } catch (PolicyServiceNotFoundException e) {
+      // Not found is not an error as far as BPM is concerned.
```
aherbst-broad, line 82: Finally grokking this, this is if the PAO is not found which makes sense here. The nested try's are a bit awkward.

service/src/main/resources/application.yml
```diff
@@ -77,6 +79,10 @@ profile:
     base-path: ${env.sam.basePath}
     admins-group-email: ${env.sam.adminsGroupEmail}
 
+  policy:
+    client-credential-file-path: build/resources/main/generated/bpm-client-sa.json
```
aherbst-broad, line 83: Are there any helmfile changes that need to happen to get this file into place?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -558,6 +560,8 @@ components:
         createdBy:
           type: string
           description: Identity of the user who originally created the profile
+        policies:
```
aherbst-broad, line 563: 🤔 I don't think there even is an implemented update endpoint. I'd exclude this for now to avoid any confusion that updating the TPS policies is even an option.

aherbst-broad (approved):

service/build.gradle
marctalbott: There isn't unfortunately, not sure why but it seems it's a standard across our Java clients (see [Rawls deps](https://github.com/broadinstitute/rawls/blob/develop/project/Dependencies.scala#L129) for example)

service/src/main/resources/api/service_openapi.yaml
marctalbott: I think the github minimizing was confusing here, this `policies` field is for the `ProfileModel` which is the standard profile response. It's not for `UpdateProfileRequest` like it looks like. Let me know if that's not what happened and you meant something else!

aherbst-broad: Ah! Thanks.

render_configs.sh
```diff
@@ -16,4 +16,15 @@ function azure_creds {
   > $LOCAL_CONFIG_OUTPUT
 }
 
+
+function service_creds {
+  LOCAL_TPS_SA_PATH="service/src/main/resources/generated/bpm-client-sa.json"
+  TPS_VAULT_PATH="secret/dsde/terra/kernel/dev/dev/bpm/app-sa"
```
cahrens, line 22: Is this path correct? (dev/dev/)

service/src/main/resources/application.yml
marctalbott: I forgot the values in this config _are_ for local runs. Helmfile PR is here - https://github.com/broadinstitute/terra-helmfile/pull/4764

service/src/main/java/bio/terra/profile/service/policy/TpsApiDispatch.java
marctalbott: Reworded the comment a bit to hopefully help with that a bit. Agreed that the nested try's are awkward, but I'd rather deal with a converted ApiException (`PolicyServiceNotFoundException`) and not have to duplicate the code in  `convertApiException` that catches 404s

blakery (approved):

PR #324, opened 2023-11-15
QA-2177: Regulate the jobs that use bumper action, to prevent tag dispatch during workflows other than "Tag, publish, deploy".

ichengchang:
Ticket: [QA-2177](https://broadworkbench.atlassian.net/browse/QA-2177)

To prevent bumper action to dispatch `semver` tag more than once on PR merge.

[QA-2177]: https://broadworkbench.atlassian.net/browse/QA-2177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/consumer_contract_tests.yml
+++ b/.github/workflows/consumer_contract_tests.yml
@@ -122,7 +122,7 @@ jobs:
             "repo-owner": "${{ github.repository_owner }}",
             "repo-name": "${{ github.event.repository.name }}",
             "repo-branch": "${{ needs.init-github-context.outputs.repo-branch }}" ,
-            "release-tag": "${{ needs.bpm-app-version.outputs.tag }}"
+            "release-tag": "${{ needs.bpm-app-version.outputs.new-tag }}"
           }'
 
 ## Deactivated until WOR-890 lands
@@ -142,5 +142,5 @@ jobs:
 #          inputs: '{
 #            "run-name": "${{ env.CAN_I_DEPLOY_RUN_NAME }}",
 #            "pacticipant": "bpm",
-#            "version": "${{ needs.bpm-app-version.outputs.tag }}"
+#            "version": "${{ needs.bpm-app-version.outputs.new-tag }}"
 #          }'
--- a/.github/workflows/consumer_contract_tests.yml
+++ b/.github/workflows/consumer_contract_tests.yml
@@ -26,8 +26,28 @@ env:
   CAN_I_DEPLOY_RUN_NAME: 'can-i-deploy-${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}'
 
 jobs:
+  bump-check:
+    runs-on: ubuntu-latest
+    outputs:
+      is-bump: ${{ steps.skiptest.outputs.is-bump }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Skip version bump merges
+        id: skiptest
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
+
+  bpm-app-version:
+    needs: [ bump-check ]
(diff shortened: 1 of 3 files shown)


.github/workflows/tag.yml
```diff
@@ -3,15 +3,23 @@ on:
...
+        description: "Echo tag to console"
+        default: true
         required: false
         type: string
     outputs:
       tag:
-        description: "Generated tag"
+        description: "The value of the latest tag after running this action"
         value: ${{ jobs.tag-job.outputs.tag }}
+      new-tag:
```
marctalbott, line 20: I don't understand the difference between `new-tag` and `tag`. If `tag` is the latest tag, then wouldn't that be equivalent to `new-tag`? As it currently stands, I wouldn't know which value to use if I needed to reference a tag in a GHA

ichengchang: This is how it goes. The `new-tag` is always the tag resulting from a bump to the original tag. However, the `tag`is by definition the value of the latest tag after running the action, which might not change if dry run is used, and remains same as the original tag.

.github/workflows/consumer_contract_tests.yml
```diff
@@ -26,8 +26,42 @@ env:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Skip version bump merges
+        id: skiptest
+        uses: ./.github/actions/bump-skip
+        with:
+          event-name: ${{ github.event_name }}
+
+# We regulate the tag job to meet the following requirements according to the trigger event type:
+# 1. pull_request event (due to opening or updating of PR branch):
```
jladieu, line 42: Reading this and scratching my head in confusion...

Shouldn't tags only get created for real when a merge to main happens?

I'm getting a bit hung up `dry-run` here being used for what seems like the exact opposite of my understanding of what a dry-run is for.

IMO, anything that's on a branch and is _not_ main is considered dry-run territory, not the opposite, as used here.

jladieu: Just saw your explanation to this effect in Slack, which I think illuminates things a bit more.  I wonder if this text would help if included in the comment too?

"The key part about this PR is that the actual tag should only be dispatched during "Tag, publish, deploy" workflow, but because contract tests need to know the new tag in advance before actual deployment takes place, it relies on the bumper's dry run feature to get the next tag to publish contracts and results from verifying consumer pacts, without actually dispatching the tag. Otherwise, we need to orchestrate multiple workflows and that means a larger restructuring will need to take place. This is because Sherlock use semver instead of git hash for deployment."

ichengchang: The bumper doc [](https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper) has this definition

```
DRY_RUN (optional) - Determine the next version without tagging the branch. The workflow can use the outputs new_tag and tag in subsequent steps. Possible values are true and false (default).
```

jladieu (approved):
IMO this logic is tricky and a bit hard to reason about.

In this case I think lends itself to as much embedded explanation as possible.  I'd say overdo it on explanatory comments; in particular bring in the explanation from the PR title "to prevent tag dispatch during workflows other than "Tag, publish, deploy" and the clarification you gave in the PR comment for tag. vs new_tag

.github/workflows/consumer_contract_tests.yml
ichengchang: More comments have been added to `regulated-tag-job` section.

knapii-developments (approved):
Aproving to unblock people. Will circle back

.github/workflows/tag.yml
ichengchang: I have incorporated more comments about `tag` vs. `new-tag`.

PR #346, opened 2023-11-28
WOR-1369 [BPM] Configure dependabot to ignore some updates for client-javax module

sergiygetlin:
This pull request sets some ignore rules for dependabot to avoid updates for specific libraries. Some library updates might force us to use "jakarta" namespace instead of current "javax". This particular "client-javax" module is not supposed to use "jakarta" since this module is used by Rawls which doesn't support jakarta yet. Currently there are 2 modules which represents API client libs: 1) "client" (support jakarta) 2) client-javax (doesn't support jakarta)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,3 +19,21 @@ updates:
     ignore:
       - dependency-name: "au.com.dius.pact*"
         update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+  - package-ecosystem: "gradle"
+      directory: "/client-javax"
+      open-pull-requests-limit: 10
+      schedule:
+        interval: "weekly"
+        time: "06:00"
+        timezone: "America/New_York"
+      target-branch: "main"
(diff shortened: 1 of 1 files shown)


cahrens (approved):

.github/dependabot.yml
```diff
@@ -19,3 +19,21 @@ updates:
     ignore:
       - dependency-name: "au.com.dius.pact*"
         update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+  - package-ecosystem: "gradle"
```
aherbst-broad, line 22: Please add a comment or ensure there is a beefed up description on the PR and resulting commit msg explaining the "why" of this change so we can understand if/when we want to remove this ignore in the future.

aherbst-broad (approved):

.github/dependabot.yml
sergiygetlin: Updated PR description.

PR #356, opened 2023-12-04
Fix syntax, ignore all of jersey for client-javax

aherbst-broad:
We don't want version updates for anything jersey in client-javax.


aherbst-broad:
cc: @sergiygetlin, I think you noted this on another PR.

marctalbott (approved):

blakery (approved):

PR #374, opened 2023-12-14
[WOR-1413] Clean up our error handling during billing profile creation

aherbst-broad:
Playing around with BPM creating GCP billing profiles, I bumped into a few issues:
* Submitting a billing account ID larger than the varchar field in the DB allows leaks the details of the `INSERT` statement in the API error 😢 . I've added a global exception handler that will catch the data source exceptions and give back a more opaque internal server error
* Related, I've added validation that the billing account ID field fits in 32 chars and if not gives back an `InvalidFieldException` 
* Using a bad access token will no longer give back the slightly wonky `Error getting user email from Sam:`; instead it will return a plain "Unauthorized" message.
* Breaks out a separate handler for the mismatched arg type handling to give back more meaningful error messages (i.e., when the date for spend reporting cannot be parsed we were previously giving back a blunt error msg and the exception type, now it will output the bad parameter's name and the expected parameter type).

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/controller/GlobalExceptionHandler.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/GlobalExceptionHandler.java
@@ -7,6 +7,7 @@ import bio.terra.profile.model.ErrorReport;
 import io.sentry.Sentry;
 import jakarta.validation.ConstraintViolationException;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


blakery (approved):
Nice to see this sort of clean up.

sergiygetlin (approved):

PR #402, opened 2024-02-08
[WOR-852] Update profile

marctalbott:
Ticket: [WOR-852](https://broadworkbench.atlassian.net/browse/WOR-852)
* Implement `updateProfile` 
* Add new API to remove the billing account from a billing profile
    * This could be modeled by passing in an empty string for the billing account id, but a dedicated API seemed clearer. Let me know if you disagree!

[WOR-852]: https://broadworkbench.atlassian.net/browse/WOR-852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/main/java/bio/terra/profile/service/iam/model/SamAction.java
```diff
@@ -8,6 +8,7 @@ public enum SamAction {
   CREATE("create"),
   DELETE("delete"),
   UPDATE_BILLING_ACCOUNT("update_billing_account"),
+  UPDATE_METADATA("update_metadata"),
```
marctalbott, line 11: This action is granted to `spend-profile` owners ([source](https://github.com/broadinstitute/sam/blob/b46ad2fc1a2cd1959a4e2623fcc5e8b63dc19c05/src/main/resources/reference.conf#L1100))

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -418,13 +436,7 @@ components:
 
     UpdateProfileRequest:
       type: object
-      required:
-        - id
```
marctalbott, line 422: `id` is included as a path parameter, so removed it from the request body.

service/src/test/java/bio/terra/profile/service/profile/flight/update/UpdateProfileFlightTest.java
```diff
@@ -0,0 +1,231 @@
...
+        .verifyAuthorization(
+            userRequest, SamResourceType.PROFILE, profile.id(), SamAction.UPDATE_BILLING_ACCOUNT);
+    verify(samService)
+        .verifyAuthorization(
+            userRequest, SamResourceType.PROFILE, profile.id(), SamAction.UPDATE_METADATA);
+    verify(billingCow).testIamPermissions(any());
+  }
+
+  @Test
+  void updateProfileBillingAccountSuccess() throws InterruptedException {
```
marctalbott, line 112: This test and `updateProfileDescriptionSuccess` are arguably not worth including. We're already testing that the flight is properly constructed in `UpdateProfileFlightUnitTest` depending on the update request body. 

The one benefit IMO is these tests would catch a bug where the overall flight's success becomes dependent on one of the steps that is not required for the flight (e.g. `VerifyUserBillingAccountAccessStep` adds a required value to the working map that won't be present if the flight is only updating the profile description).

service/src/test/java/bio/terra/profile/service/profile/flight/update/UpdateProfileFlightTest.java
```diff
@@ -0,0 +1,231 @@
...
+        .verifyAuthorization(
+            userRequest, SamResourceType.PROFILE, profile.id(), SamAction.UPDATE_METADATA);
+    verify(samService, times(0))
+        .verifyAuthorization(
+            userRequest, SamResourceType.PROFILE, profile.id(), SamAction.UPDATE_BILLING_ACCOUNT);
+    verifyNoInteractions(crlService);
+  }
+
+  @Test
+  void updateProfileUnauthorizedFailure() throws InterruptedException {
```
marctalbott, line 207: I also opted not to test every possible authz check failure, but could be convinced to add them in

service/src/main/java/bio/terra/profile/service/profile/flight/update/UpdateProfileRecordStep.java
```diff
@@ -0,0 +1,38 @@
...
+    @Nullable String billingAccountId)
+    implements Step {
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    if (profileDao.updateProfile(profile.id(), description, billingAccountId)) {
+      return StepResult.getStepResultSuccess();
+    } else {
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
```
blakery, line 25: Could this fail for something non-fatal, like a temporary database connection issue?

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
```diff
@@ -167,6 +162,32 @@ public List<ProfileDescription> listProfiles(
...
+            .newJob()
+            .description(description)
+            .flightClass(UpdateProfileFlight.class)
+            .request(requestBody)
+            .addParameter(ProfileMapKeys.PROFILE, profile)
+            .userRequest(user);
+    return updateJob.submitAndWait(BillingProfile.class).toApiProfileModel();
+  }
+
+  public void removeBillingAccount(UUID id, AuthenticatedUserRequest user) {
```
blakery, line 181: Why does the normal profile update require stairway, but `removeBillingAccount` doesn't?

On a related note, I wonder if we might be over-using stairway, both here, and maybe even more generally. 
I'm not sure we really need the multi-step, all-or-nothing, unwindable guarantees when there's only a single change being made, and it's to this service's db. Unless I'm missing something, the rest of the steps are just validations and access checks, which there's no reason to run in stairway. 

I guess there's the separate part of returning the updated object, but we could either not do that, or find some other way that doesn't require a separate db call (looks like `ProfileDao.createBillingProfile()` already just returns the pojo from memory, instead of running a separate query).

blakery (commented):
I also prefer the separate endpoint. It's too easy for passing in an empty string to either look like a mistake, or for the behavior to be changed accidentally, etc.

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
marctalbott: Yeah, I went back and forth on whether to make this a flight or not. Ultimately, I landed on the flight when introducing the GCP authz check as the external cloud call pushed me over the edge. It does give us a bit more flexibility going forward in case we need more complex profile updates, but I wouldn't say that alone is worth the added complexity of a flight.

As for your original question, a single authz check with Sam followed by the update didn't feel worth a flight. Perhaps that's telling enough for the original flight question -- adding two authz checks are probably not enough to warrant a flight

service/src/main/java/bio/terra/profile/service/profile/flight/update/UpdateProfileRecordStep.java
marctalbott: I expect that would result in `updateProfile` throwing an uncaught (😢) exception. `updateProfile` will only return false if no profile was found, in which case we should bail out of the flight IMO

service/src/main/java/bio/terra/profile/service/gcp/GcpService.java
```diff
@@ -0,0 +1,63 @@
...
+      List.of("billing.resourceAssociations.create");
+
+  private final GcpCrlService crlService;
+
+  @Autowired
+  public GcpService(GcpCrlService crlService) {
+    this.crlService = crlService;
+  }
+
+  public void verifyUserBillingAccountAccess(
```
marctalbott, line 33: This was mostly pulled from `CreateProfileVerifyAccountStep`. The major changes were around error handling

service/src/main/java/bio/terra/profile/db/ProfileDao.java
```diff
@@ -156,6 +159,45 @@ public boolean deleteBillingProfileById(UUID id) {
...
+      params.addValue("billing_account_id", billingAccountId);
+      setClause.add("billing_account_id = :billing_account_id");
+    }
+
+    setClause.add("last_modified = current_timestamp(6)");
+
+    String sql =
+        String.format("UPDATE billing_profile SET %s WHERE id = :id", String.join(",", setClause));
+
+    int rowsAffected = jdbcTemplate.update(sql, params);
```
sergiygetlin, line 188: probably can delete rowsAffected.

service/src/main/java/bio/terra/profile/service/gcp/GcpService.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+public class GcpService {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(bio.terra.profile.service.gcp.GcpService.class);
+
+  public static final List<String> BILLING_ACCOUNT_PERMISSIONS_TO_TEST =
+      List.of("billing.resourceAssociations.create");
```
sergiygetlin, line 24: If it's always just one value maybe can use Collections.singletonList() instead.

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlight.java
```diff
@@ -39,7 +39,9 @@ public CreateProfileFlight(FlightMap inputParameters, Object applicationContext)
     addStep(new CreateProfileStep(profileDao, billingProfile, user));
     switch (billingProfile.cloudPlatform()) {
       case GCP:
-        addStep(new CreateProfileVerifyAccountStep(crlService, billingProfile, user));
+        addStep(
```
sergiygetlin, line 42: Not sure I understand. Why are we changing create profile logic? How this is related to update profile?

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -152,6 +152,24 @@ paths:
         '500':
           $ref: '#/components/responses/ServerError'
 
+  /api/profiles/v1/{profileId}/billingAccount:
```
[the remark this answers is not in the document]
marctalbott: Correct, there can only be one billing account related to a billing profile

service/src/main/java/bio/terra/profile/app/controller/ProfileApiController.java
```diff
@@ -60,13 +60,27 @@ public ResponseEntity<ProfileModelList> listProfiles(Integer offset, Integer lim
     return new ResponseEntity<>(response, HttpStatus.OK);
   }
 
+  @Override
```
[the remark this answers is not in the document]
marctalbott: It was defined in the openapi spec, but not implemented. If you try to update a profile in dev, it will return a 501

service/src/main/java/bio/terra/profile/db/ProfileDao.java
```diff
@@ -156,6 +159,45 @@ public boolean deleteBillingProfileById(UUID id) {
     }
   }
 
+  @WriteTransaction
+  public boolean updateProfile(
+      UUID id, @Nullable String description, @Nullable String billingAccountId) {
```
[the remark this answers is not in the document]
marctalbott: Yes, at least one field needs to be updated.

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlight.java
marctalbott: I renamed the step when there was a separate `UpdateProfileFlight`. Now that there isn't a separate `UpdateProfileFlight`, I'll revert the name change.

The step itself was changed to use the new `GcpService` so we can reuse code in the step and in the `ProfileService`.

service/src/main/java/bio/terra/profile/service/gcp/GcpService.java
marctalbott: 👍 I'll check that out. This was pulled out of the old step, but happy to make that change

blakery (approved):

sergiygetlin (approved):

PR #418, opened 2024-03-18
[WOR-1560] Upgrade to TCL 1.1.1 for MDC-aware Stairway

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1560

Upgraded `terra-common-lib` from 0.1.6 -> 1.1.1 mainly for the benefit of [Stairway 1.0.0's native MDC-awareness](https://github.com/DataBiosphere/stairway/pull/125).

As a result, MdcHook is removed in favor of TCL's common [StairwayLoggingHook](https://github.com/DataBiosphere/terra-common-lib/pull/165), which only logs at notable state transitions.

BPM now also gets more context than it previously had. Previously, it only logged the context of the calling thread for logs emitted during the execution of each step. Now, it…

- logs the context of the calling thread, along with generic flight context (flight class and flight ID) for the duration of the flight's execution on its new thread
- additionally logs generic step context (step class, index, and direction) for the duration of each step's execution

This matches what's been live in WSM for awhile.

**Dependabot for Spring Boot upgrades**

One of the most impactful things we can do to stay on top of vulnerable dependencies is to keep Spring Boot up to date.  We use Spring's dependency manager and pull in many of its starter packages as dependencies, so as Spring Boot addresses vulnerabilities on its ends we'd like to get them for free, rather than be tempted to pin versions solely for the purpose of mitigating a vulnerability.

So, I've adjusted our Dependabot settings to stop ignoring Spring Boot updates. 

**Running SourceClear scans locally**

I added Gradle's srcclr plugin and instructions for local scans, following examples set by other repos.  Given the extent of dependency modifications in this PR, I appreciated having a way to check locally that I was both addressing vulnerabilities, and not pulling in any new ones.

Some other teams also run SourceClear as part of their CI build process, but I am not making that change at this time.

**Other dependency changes**

I upgraded Spring Boot to latest, 3.2.3, and cleaned up dependencies and pins no longer needed.  This upgrade will address a number of vulnerabilities flagged by Veracode, confirmed by running SourceClear locally off of this feature branch.

OpenTelemetry direct dependencies could be drastically reduced, but we do have to make sure that Spring Boot dependency manager is bringing in a version of `opentelemetry-bom` compatible with what TCL is using.


README.md
```diff
@@ -76,5 +76,23 @@ Automatically fix linting issues:
 ./gradlew generateSwaggerCode
 ```
 
+### Running SourceClear locally
```
cahrens, line 79: Thank you for documenting this!!

okotsopoulos: My pleasure! I will have a follow-on to get this in LZS too, and will get it in WSM.

buildSrc/src/main/groovy/bio.terra.profile.java-common-conventions.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
 
     id 'com.diffplug.spotless'
     id 'org.hidetake.swagger.generator'
+    id 'com.srcclr.gradle'
```
sergiygetlin, line 8: Is this tool used across different project in Terra? Is it something new or BPM haven't used it yet?

sergiygetlin (approved):

cahrens (approved):

buildSrc/src/main/groovy/bio.terra.profile.java-common-conventions.gradle
okotsopoulos: It is used in a number of Terra projects, but it's wired in inconsistently. 

Here are a few examples: https://github.com/search?q=org%3ADataBiosphere+com.srcclr.gradle%3Acom.srcclr.gradle.gradle.plugin&type=code

Here's an example of a repo that has Gradle's srcclr plug-in enabled, and also runs the scan as part of their builds:
https://github.com/search?q=repo%3ADataBiosphere%2Fjava-pfb%20srcclr&type=code

LZS actually pulls in the plug-in, but doesn't currently use it anywhere.

PR #419, opened 2024-03-19
[TOAZ-343]: api/azure/v1/managedApps support for ServiceCatalog deployed azure managed apps

aaronegrant:
- Uses the azure-control-plane-enabled feature flag to switch between finding Marketplace/ServiceCatalog deployed apps
- For ServiceCatlog deployed apps, we don't use the publisher or plan AMA properties, instead we are looking at the kind (''ServiceCatalog") and matching up the authuser email
- New test for ServiceCatalog app kind

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/configuration/PolicyServiceConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/PolicyServiceConfiguration.java
@@ -9,6 +9,7 @@ import com.google.auth.oauth2.ServiceAccountCredentials;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
@@ -21,7 +22,7 @@ public class PolicyServiceConfiguration {
 
   private String basePath;
   private String clientCredentialFilePath;
-  private Boolean azureControlPlaneEnabled;
+  private final AzureConfiguration azureConfiguration;
 
   private static final List<String> POLICY_SERVICE_ACCOUNT_SCOPES =
       List.of("openid", "email", "profile");
@@ -42,16 +43,13 @@ public class PolicyServiceConfiguration {
     return clientCredentialFilePath;
   }
 
-  public void setAzureControlPlaneEnabled(Boolean azureControlPlaneEnabled) {
-    this.azureControlPlaneEnabled = azureControlPlaneEnabled;
-  }
-
-  public Boolean getAzureControlPlaneEnabled() {
-    return azureControlPlaneEnabled;
+  @Autowired
+  PolicyServiceConfiguration(AzureConfiguration azureConfiguration) {
+    this.azureConfiguration = azureConfiguration;
   }
 
   public String getAccessToken() throws IOException {
-    if (azureControlPlaneEnabled) {
+    if (azureConfiguration.controlPlaneEnabled()) {
       TokenCredential credential = new DefaultAzureCredentialBuilder().build();
       // The Microsoft Authentication Library (MSAL) currently specifies offline_access, openid,
       // profile, and email by default in authorization and token requests.
--- a/service/src/main/java/bio/terra/profile/app/controller/AzureApiController.java
+++ b/service/src/main/java/bio/terra/profile/app/controller/AzureApiController.java
@@ -7,7 +7,6 @@ import bio.terra.profile.model.AzureManagedAppModel;
 import bio.terra.profile.model.AzureManagedAppsResponseModel;
 import bio.terra.profile.service.azure.AzureService;
 import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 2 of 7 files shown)


service/src/main/java/bio/terra/profile/app/controller/AzureApiController.java
```diff
@@ -18,24 +21,38 @@ public class AzureApiController implements AzureApi {
...
   public ResponseEntity<AzureManagedAppsResponseModel> getManagedAppDeployments(
       UUID azureSubscriptionId, Boolean includeAssignedApplications) {
     final AuthenticatedUserRequest userRequest = authenticatedUserRequestFactory.from(request);
-    var managedApps =
-        this.azureService.getAuthorizedManagedAppDeployments(
-            azureSubscriptionId, includeAssignedApplications, userRequest);
+    List<AzureManagedAppModel> managedApps;
+
+    // check azure control plane feature flag
+    if (this.policyServiceConfiguration.getAzureControlPlaneEnabled()) {
```
dvoet, line 45: it does not seem like `PolicyServiceConfiguration` is the right place for the config setting `azureControlPlaneEnabled`

aaronegrant: @marctalbott @sergiygetlin any suggestions on where to put the config for the AzureControlPlane flag?

sergiygetlin: ExternalConfiguration?

marctalbott: `AzureConfiguration` would likely be the best existing location for it. Or you could set up a new configuration class like `AzureControlPlaneConfiguration` specifically for control plane-related config. Either is fine with me

service/src/main/java/bio/terra/profile/app/controller/AzureApiController.java
```diff
@@ -18,24 +21,38 @@ public class AzureApiController implements AzureApi {
...
+      // get ServiceCatalog deployed managed apps
+      managedApps =
+          this.azureService.getServiceCatalogManagedAppDeployments(
+              azureSubscriptionId, includeAssignedApplications, userRequest);
+    } else {
+      // get Marketplace deployed managed apps
+      managedApps =
+          this.azureService.getAuthorizedManagedAppDeployments(
+              azureSubscriptionId, includeAssignedApplications, userRequest);
+    }
```
marctalbott, line 55: I am wondering if this would be simpler if it diverged in `AzureService`. `getServiceCatalogManagedAppDeployments` and `getAuthorizedManagedAppDeployments` are almost identical methods, the only difference that I'm seeing is one filters based on `isServiceCatalogTerraManagedApp` and the other uses `isAuthedTerraManagedApp`

aaronegrant: We can push the switch down into AzureService isAuthedTerraManagedApp(). I had it there initially, but thought this way brings the feature flag up to a level that shows we have different behavior for a single API. Take a look at AzureService in cancelled PR #415 and let me know if you prefer that approach

service/src/main/java/bio/terra/profile/app/controller/AzureApiController.java
sergiygetlin: Should we update AzureApiController test as well since original flow splits into two parts here (line 45) or per Marc's suggestion maybe we should move it to service level and this is where this split is controlled via specific mocking in the test - `when(policyServiceConfiguration.getAzureControlPlaneEnabled()).thenReturn(true|false)`.

service/src/main/java/bio/terra/profile/app/controller/AzureApiController.java
dvoet: I agree the controller is not the right place. Controllers should have little to no business logic.

service/src/main/java/bio/terra/profile/app/configuration/PolicyServiceConfiguration.java
```diff
@@ -42,16 +43,13 @@ public String getClientCredentialFilePath() {
...
-  public Boolean getAzureControlPlaneEnabled() {
-    return azureControlPlaneEnabled;
+  @Autowired
+  PolicyServiceConfiguration(AzureConfiguration azureConfiguration) {
+    this.azureConfiguration = azureConfiguration;
   }
 
   public String getAccessToken() throws IOException {
-    if (azureControlPlaneEnabled) {
+    if (azureConfiguration.controlPlaneEnabled()) {
```
sergiygetlin, line 52: Now I see why the controlPlaneEnabled was originally here. This is probably out of scope for this PR, but I think it would be better not to inject one configuration into another. We might need to put this getAccessToken into separate utility so we can inject (or even pass as a param) only AzureConfiguration instead. To be clear what I mean is to 1) move getAccessToken into utility class 2) Inject AzureConfiguration into TpsApiDispatch and pass required flag to generate token. @marctalbott what do you think?

marctalbott (approved):
There's still some refactoring that could be done around the `PolicyServiceConfiguration`, but doesn't have to happen in this PR

sergiygetlin (approved):

PR #420, opened 2024-03-20
[WOR-1562] Use Google access token for Google calls

marctalbott:
Ticket: [WOR-1562](https://broadworkbench.atlassian.net/browse/WOR-1562)
* We need to use a Google access token to for Google API calls. The token in the `AuthenticatedUserRequest` can be either a B2C JWT or a Google access token. If the token is a B2C JWT, it will have the user's Google access token included as a `idp_access_token` claim which we need to use for Google API calls. If the token isn't a JWT, we assume it's a valid Google access token and attempt to use it anyways.

[WOR-1562]: https://broadworkbench.atlassian.net/browse/WOR-1562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/crl/GcpCrlService.java
+++ b/service/src/main/java/bio/terra/profile/service/crl/GcpCrlService.java
@@ -6,7 +6,9 @@ import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.profile.service.crl.exception.CrlInternalException;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.nimbusds.jwt.SignedJWT;
 import java.io.IOException;
+import java.text.ParseException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -15,6 +17,7 @@ import org.springframework.stereotype.Component;
 public class GcpCrlService {
 
   private final ClientConfig clientConfig;
+  private static final String GOOGLE_ACCESS_TOKEN_CLAIM = "idp_access_token";
 
   @Autowired
   public GcpCrlService(ClientConfig clientConfig) {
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/profile/service/crl/GcpCrlService.java
```diff
@@ -33,4 +37,14 @@ public CloudBillingClientCow getBillingClientCow(AuthenticatedUserRequest user)
   private GoogleCredentials getUserCredentials(String token) {
     return GoogleCredentials.newBuilder().setAccessToken(new AccessToken(token, null)).build();
   }
+
+  private String getGoogleAccessToken(String userToken) {
+    try {
+      var jwt = SignedJWT.parse(userToken);
+      return (String) jwt.getJWTClaimsSet().getClaim(GOOGLE_ACCESS_TOKEN_CLAIM);
```
dvoet, line 44: I generally stay away from parsing the JWT since infosec cares very much that you do it right. This may be checking the signature, but it isn't checking the issuer or audience or other things they may care about.

Did you know that the apache proxy extracts all claims in the B2C JWT and adds them to request headers prefixed with `OAUTH2_CLAIM_`. You can find what you need in the `OAUTH2_CLAIM_idp_access_token` request header instead of parsing the JWT yourself.

marctalbott: BPM doesn't run with a proxy when it runs locally which was the motivation for doing it this way. Maybe it would be better to only get the idp access token in this fashion if BPM is running locally and rely on the request header in all other cases.

My hope was that the user's token has been properly vetted by the time we get to this point and there was no harm in trying to grab this claim if it existed, but I will see what infosec has to say about that.

dvoet: disregard this comment

cahrens (approved):

sergiygetlin (approved):

PR #435, opened 2024-04-08
[TOAZ-346] Use workload Managed Identity for ARM authentication when running under AzureControlPlane

aaronegrant:
Use DefaultAzureCredentialBuilder() with the service workload client Id when running under AzureControlPlane


service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
```diff
@@ -63,11 +65,17 @@ public void setAuthorizedUserKey(String authorizedUserKey) {
   }
 
   public TokenCredential buildManagedAppCredentials() {
-    return new ClientSecretCredentialBuilder()
-        .clientId(managedAppClientId)
-        .clientSecret(managedAppClientSecret)
-        .tenantId(managedAppTenantId)
-        .build();
+    if (controlPlaneEnabled) {
```
sergiygetlin, line 68: I think we might need test for this public method to ensure that different CredentialBuilders used.

service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
```diff
@@ -12,6 +13,7 @@ public record AzureConfiguration(
     String managedAppClientId,
     String managedAppClientSecret,
     String managedAppTenantId,
+    String managedAppWorkloadClientId,
     Boolean controlPlaneEnabled,
```
sergiygetlin, line 17: SonarCloud also suggested using primitive type here if it's possible.

service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
aaronegrant: @sergiygetlin do you have an example of an existing test that I should look at? Seems like we use mocks for anything that would require a credential

sergiygetlin: @aaronegrant I would try to split configuration and TokenCredential generation. Making TokenCredentialFactory and passing required boolean into the factory method.

aaronegrant: @sergiygetlin refactored to used ChainedTokenCredentialBuilder. Will first try to use a ManagedIdentity, and if that is not succesful, will fall through to use the historical ServicePrincipal auth. Creating a test for the MI auth path isn't doable since that auth method only works when running BPM in an Azure service. See the note in blue at https://learn.microsoft.com/en-us/azure/azure-app-configuration/howto-integrate-azure-managed-service-identity?tabs=core5x&pivots=framework-dotnet

service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
```diff
@@ -62,11 +62,27 @@ public void setAuthorizedUserKey(String authorizedUserKey) {
     }
   }
 
-  public TokenCredential buildManagedAppCredentials() {
-    return new ClientSecretCredentialBuilder()
-        .clientId(managedAppClientId)
-        .clientSecret(managedAppClientSecret)
-        .tenantId(managedAppTenantId)
+  public ChainedTokenCredential createChainedCredential() {
```
dvoet, line 65: This should return `TokenCredential` and be called `createTokenCredential`, the fact that it is chained is an implementation detail

service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
dvoet: I agree that I don't see an adequate way to test this in a unit test. In the first iteration of this code, when there was a conditional, I could see a test that ensures the correct code path executes, but I think that is irrelevant when using the chained credential approach.

sergiygetlin (approved):

blakery (approved):

PR #436, opened 2024-04-09
[WOR-1580] Add organization to ProfileModel

marctalbott:
Ticket: [WOR-1580](https://broadworkbench.atlassian.net/browse/WOR-1580)
* When returning a billing profile, check its subscription against the allowlist of enterprise subscriptions to determine if it's an enterprise billing profile

[WOR-1580]: https://broadworkbench.atlassian.net/browse/WOR-1580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/profile/model/ProfileDescription.java
+++ b/service/src/main/java/bio/terra/profile/service/profile/model/ProfileDescription.java
@@ -2,31 +2,36 @@ package bio.terra.profile.service.profile.model;
 
 import bio.terra.policy.model.TpsPolicyInputs;
 import bio.terra.profile.model.CreateProfileRequest;
+import bio.terra.profile.model.Organization;
 import bio.terra.profile.model.ProfileModel;
 import bio.terra.profile.service.policy.TpsConversionUtils;
 import java.util.Objects;
 import java.util.Optional;
 
 public record ProfileDescription(
-    BillingProfile billingProfile, Optional<TpsPolicyInputs> policies) {
+    BillingProfile billingProfile,
+    Optional<TpsPolicyInputs> policies,
+    Optional<Organization> organization) {
   public ProfileDescription {
     Objects.requireNonNull(billingProfile);
     Objects.requireNonNull(policies);
   }
 
   public ProfileDescription(BillingProfile billingProfile) {
-    this(billingProfile, Optional.empty());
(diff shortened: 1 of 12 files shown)


service/src/main/java/bio/terra/profile/service/profile/model/ProfileDescription.java
```diff
@@ -2,31 +2,36 @@
...
-    this(billingProfile, Optional.empty());
+    this(billingProfile, Optional.empty(), Optional.empty());
   }
 
   public static ProfileDescription fromApiCreateProfileRequest(CreateProfileRequest request) {
     return new ProfileDescription(
         BillingProfile.fromApiCreateProfileRequest(request),
-        Optional.ofNullable(TpsConversionUtils.tpsFromBpmApiPolicyInputs(request.getPolicies())));
+        Optional.ofNullable(TpsConversionUtils.tpsFromBpmApiPolicyInputs(request.getPolicies())),
+        Optional.empty());
```
cahrens, line 28: Don't we want to try to figure out of if the subscription is in the enterprise list here?

marctalbott: It's a bit awkward to do here since we don't have access to the `EnterpriseConfiguration` needed to determine if the `CreateProfileRequest` will be an enterprise organization. I'm looking into refactoring `ProfileDescription` to see if I can make it a little easier

marctalbott: I haven't been able to refactor this into something better. I'm inclined to leave it as is since the `CreateProfileRequest` doesn't contain any organization information as it is. The `CreateProfileFlight` has been changed to return the organization information after creating the profile

marctalbott: I am not sure if a `record` with multiple `Optional` fields is the best way to represent the `ProfileDescription` or if there's a simpler way to handle the `EnterpriseConfiguration`, but I haven't been able to come up with better alternatives. Open to suggestions on this

blakery (approved):

cahrens (approved):

PR #438, opened 2024-04-11
[WOR-1632] Make azure auth token scope configurable.

jsaun:
On integration testing we found that the proxy couldn’t verify the signature of the managed identity token with scope of [graph.microsoft.com/.default](http://graph.microsoft.com/.default). This makes the scope configurable so we can set it correctly to the [client_id]/.default and change it in the future if needed for any reason with a PR.

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
@@ -13,6 +13,7 @@ public record AzureConfiguration(
     String managedAppClientSecret,
     String managedAppTenantId,
     Boolean controlPlaneEnabled,
+    String authTokenScope,
     Set<AzureApplicationOffer> applicationOffers,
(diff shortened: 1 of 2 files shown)


service/src/main/java/bio/terra/profile/app/configuration/AzureConfiguration.java
```diff
@@ -13,6 +13,7 @@ public record AzureConfiguration(
     String managedAppClientSecret,
     String managedAppTenantId,
     Boolean controlPlaneEnabled,
+    String authTokenScope,
```
sergiygetlin, line 16: Do we need default value for this?

jsaun: No, it will need to be set in helm to the client_id/.default

blakery (approved):

sergiygetlin (approved):

PR #445, opened 2024-04-30
WOR-1458 - Pact Test Fixes

blakery:
https://broadworkbench.atlassian.net/browse/WOR-1458

* Re-enable can-i-deploy for consumer contract tests (now that https://broadworkbench.atlassian.net/browse/WOR-890 is finished)
* publish all pacts in output directory to the broker

### DIFF
--- a/.github/workflows/consumer_contract_tests.yml
+++ b/.github/workflows/consumer_contract_tests.yml
@@ -165,7 +165,7 @@ jobs:
           ref: refs/heads/main
           token: ${{ secrets.BROADBOT_TOKEN }} # github token for access to kick off a job in the private repo
           inputs: '{
-            "run-name": "${{ env.PUBLISH_CONTRACTS_RUN_NAME }}-${{ matrix.pact-path }}",
+            "run-name": "${{ env.PUBLISH_CONTRACTS_RUN_NAME }}-${{ matrix.pact_path }}",
             "pact-b64": "${{ steps.encode-pact.outputs.pact-b64 }}",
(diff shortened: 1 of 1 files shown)


.github/workflows/consumer_contract_tests.yml
```diff
@@ -121,50 +122,72 @@ jobs:
...
+        with:
+          name: ${{ env.PACTS_ARTIFACT }}
+          path: ${{ env.PACTS_OUTPUT_DIR }}
+          retention-days: 1
 
   publish-contracts:
     runs-on: ubuntu-latest
     needs: [ bump-check, init-github-context, bpm-consumer-contract-tests, regulated-tag-job ]
+    strategy:
+      matrix:
```
cahrens, line 142: Why does this need matrix? Matrix has caused problems in the past with incorrect github actions runs getting linked, and I don't see that we are iterating over anything.

cahrens: Unless `pact-path` is more than a single path?

cahrens:
Can you provide a link to the pact broker showing where these got published? I'm just looking to understand more about how we know this is actually working.

.github/workflows/consumer_contract_tests.yml
cahrens: Yes, looking at the actual ticket for this change I see it is about publishing multiple pacts. So I understand now the matrix call.

cahrens (approved):
I would love for you to walk through the pact broker with us again.

marctalbott (approved):

PR #451, opened 2024-05-17
[AJ-1835] Add limits to BPM profile/organization

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1835
Adds the "limits" map to a billing profile's organization.  This will be used to identify which billing profiles have limits imposed, e.g. for Anvil Lite.  Verified in my BEE that Rawls can still communicate with BPM without needing to update its model (although we will want to do that).

Example of what the config would look like here: https://github.com/broadinstitute/terra-helmfile/pull/5564

### DIFF
--- a/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
@@ -1,8 +1,27 @@
 package bio.terra.profile.app.configuration;
 
+import java.util.Collections;
 import java.util.Map;
 import java.util.UUID;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
 
+@Configuration
 @ConfigurationProperties(prefix = "profile.limits")
-public record LimitsConfiguration(Map<UUID, Map<String, String>> subscriptions) {}
+public class LimitsConfiguration {
+    private Map<UUID, Map<String, String>> profiles;
+
+    public Map<UUID, Map<String, String>> getProfiles() {
+        return profiles;
+    }
+
+    public void setProfiles(Map<UUID, Map<String, String>> profiles) {
+        this.profiles = profiles;
+    }
+
+    public Map<String, String> getLimitsForProfile(UUID profileId){
+        return this.profiles.getOrDefault(profileId, Collections.emptyMap());
+    }
+}
+
+
--- a/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
@@ -2,7 +2,8 @@ package bio.terra.profile.app.configuration;
 
 import java.util.Map;
 import java.util.UUID;
+
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties(prefix = "profile.limits")
-public record LimitsConfiguration(Map<UUID, Map<String, String>> limits) {}
+public record LimitsConfiguration(Map<UUID, Map<String, String>> subscriptions) {}
--- a/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
@@ -1,11 +1,8 @@
 package bio.terra.profile.app.configuration;
 
+import java.util.Map;
+import java.util.UUID;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
-
 @ConfigurationProperties(prefix = "profile.limits")
-public record LimitsConfiguration(Map<UUID, Map<String,String>> limits) {
-}
+public record LimitsConfiguration(Map<UUID, Map<String, String>> limits) {}
--- a/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
@@ -2,7 +2,6 @@ package bio.terra.profile.app.configuration;
 
 import java.util.Map;
 import java.util.UUID;
-
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 @ConfigurationProperties(prefix = "profile.limits")
--- a/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
+++ b/service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
@@ -1,0 +1,11 @@
+package bio.terra.profile.app.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+@ConfigurationProperties(prefix = "profile.limits")
+public record LimitsConfiguration(Map<UUID, Map<String,String>> limits) {
+}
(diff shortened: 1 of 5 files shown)


service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
```diff
@@ -233,7 +238,15 @@ private Organization getProfileOrganization(BillingProfile profile) {
...
-                .orElse(false));
+                .orElse(false))
+        .limits(
+            profile
+                .subscriptionId()
+                .flatMap(
+                    id ->
+                        Optional.ofNullable(limitsConfiguration.subscriptions())
+                            .map(subscriptions -> subscriptions.get(id)))
+                .orElse(Map.of()));
```
nawatts, line 249: Could this be moved into `LimitsConfiguration` and exposed via a `getLimitsForSubscription` method or something like that?

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
```diff
@@ -122,7 +123,9 @@ void before() {
             "creator");
     profileDescription =
         new ProfileDescription(
-            profile, Optional.empty(), Optional.of(new Organization().enterprise(false)));
+            profile,
+            Optional.empty(),
+            Optional.of(new Organization().enterprise(false).limits(Map.of())));
```
nawatts, line 128: Super nit: `emptyMap` might be _slightly_ clearer.

```suggestion
            Optional.of(new Organization().enterprise(false).limits(emptyMap())));
```

nawatts: Or... could `Organization` default to an empty `limits` map so that this isn't necessary everywhere an `Organization` is created?

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
```diff
@@ -236,6 +241,38 @@ void getProfileWithEnterpriseOrganization() throws InterruptedException {
...
+
+    Object limitsObject = limitedResult.organization().get().getLimits();
+    assertTrue(limitsObject instanceof Map);
+    Map<String, String> limitsMap = (Map<String, String>) limitsObject;
+    assertTrue(limitsMap.containsKey("vcpus"));
+
+    Object nonlimitedObject = nonLimitedResult.organization().get().getLimits();
+    assertTrue(nonlimitedObject instanceof Map);
+    Map<String, String> emptyMap = (Map<String, String>) nonlimitedObject;
+    assert (emptyMap.isEmpty());
```
nawatts, line 273: Are plain `assert`s ok in tests? Or does this need to be some kind of JUnit matcher?

nawatts (approved):

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
calypsomatic: `Organization` is generated from the swagger code and I haven't found a way to make it default to an empty map instead of null.

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
aherbst-broad: I'd suggest using a junit matcher like `assertThat` rather than a raw assert.

service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.profile.app.configuration;
+
+import java.util.Map;
+import java.util.UUID;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "profile.limits")
+public record LimitsConfiguration(Map<UUID, Map<String, String>> subscriptions) {}
```
aherbst-broad, line 8: Can you add a brief header comment describing the purpose of this configuration and its intended use?

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
```diff
@@ -236,6 +241,38 @@ void getProfileWithEnterpriseOrganization() throws InterruptedException {
...
+    when(profileDao.getBillingProfileById(limitedProfile.id())).thenReturn(limitedProfile);
+    when(profileDao.getBillingProfileById(nonLimitedProfile.id())).thenReturn(nonLimitedProfile);
+    when(samService.hasActions(eq(user), eq(SamResourceType.PROFILE), any())).thenReturn(true);
+    when(tpsApiDispatch.getOrCreatePao(any(), any(), any())).thenReturn(new TpsPaoGetResult());
+
+    var limitedResult = profileService.getProfile(limitedProfile.id(), user);
+    var nonLimitedResult = profileService.getProfile(nonLimitedProfile.id(), user);
+
+    Object limitsObject = limitedResult.organization().get().getLimits();
+    assertTrue(limitsObject instanceof Map);
```
aherbst-broad, line 266: This assertion seems redundant, how would getLimits come back as anything other than a Map (or null)? You might be able to condense L265-268 by making `limitsObject` a var and leaving the assertion on L268.

service/src/test/java/bio/terra/profile/service/profile/ProfileServiceUnitTest.java
```diff
@@ -236,6 +241,38 @@ void getProfileWithEnterpriseOrganization() throws InterruptedException {
...
+    Map<String, String> limitsMap = (Map<String, String>) limitsObject;
+    assertTrue(limitsMap.containsKey("vcpus"));
+
+    Object nonlimitedObject = nonLimitedResult.organization().get().getLimits();
+    assertTrue(nonlimitedObject instanceof Map);
```
aherbst-broad, line 271: Same question about asserting the type here.

aherbst-broad (approved):

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
```diff
@@ -233,7 +238,15 @@ private Organization getProfileOrganization(BillingProfile profile) {
             profile
                 .subscriptionId()
                 .map(enterpriseConfiguration.subscriptions()::contains)
-                .orElse(false));
+                .orElse(false))
+        .limits(
+            profile
+                .subscriptionId()
+                .flatMap(
```
davidangb, line 245: I don't think `profile.subscriptionId()` is correct - this would look up the subscriptionId in the terra-helmfile config. Instead, I believe we want to be looking up the billing profile id, which would be `profile.id()`.

In other words, we are applying limits to specific billing profiles, not to entire subscriptions.

Maybe this means "limits" should be moved out from under "organization" in the response? Hmmm …

service/src/main/java/bio/terra/profile/app/configuration/LimitsConfiguration.java
davidangb, line 8: see my other comment - I think this should be "profiles", not "subscriptions"

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
calypsomatic: Spring is yelling at me that a `ConfigurationProperties` has to be a record, which means I can't add additional methods to it.  Am struggling with making config that's not a record, it might be possible!

davidangb (changes requested):

service/src/main/java/bio/terra/profile/service/profile/ProfileService.java
nawatts: Do we need the extra granularity? Or is using Organization / subscription sufficient for AnVIL Lite and trial accounts?

davidangb (approved):
lgtm, thank you! Did not test it myself.

aherbst-broad (approved):

davidangb (approved):

PR #464, opened 2024-07-11
WOR-1769 getProfile: check read_profile action instead allowing any action

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1769

see also https://github.com/broadinstitute/sam/pull/1489

### DIFF
--- a/service/src/main/java/bio/terra/profile/service/iam/SamService.java
+++ b/service/src/main/java/bio/terra/profile/service/iam/SamService.java
@@ -111,34 +111,6 @@ public class SamService {
   }
 
   /**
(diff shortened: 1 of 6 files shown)


marctalbott (approved):

calypsomatic (commented):
Does this need to be applied to `listProfiles` as well?

dvoet:
> Does this need to be applied to `listProfiles` as well?

No, `listProfiles` does not check a particular action. It uses the `listResourcesAndPoliciesV2` sam api which lists all profiles a user has some kind of access to and what that access is.

PR #474, opened 2024-08-13
WOR-865: add changelog table

blakery:
[WOR-865](https://broadworkbench.atlassian.net/browse/WOR-865)

Create changelog table and record billing profile updates to it.
When creating or deleting a billing project, entries are recorded in a stairway step.
For updates, including removing a billing account, the entries are added after the update in `ProfileService`.

[WOR-865]: https://broadworkbench.atlassian.net/browse/WOR-865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service/src/test/java/bio/terra/profile/db/ProfileDaoTest.java
blakery, line 1: Moved to correct package

service/src/main/java/bio/terra/profile/db/ProfileChangeLogDao.java
```diff
@@ -0,0 +1,154 @@
...
+    var sql =
+        INSERT_INTO_TABLE
+            + " (profile_id, change_type, change_by, changes) VALUES "
+            + "(:profile_id, :change_type, :change_by, :changes::jsonb)";
+
+    String serializedChanges = null;
+    try {
+      serializedChanges = objectMapper.writeValueAsString(changes);
+    } catch (JsonProcessingException e) {
+      logger.error(
```
aherbst-broad, line 90: 🤔 I think this is ok; if it can't serialize the changes, it logs an error and at least records an entry with a null value for the changes so there is at least some evidence in the table. I was pondering whether this should throw in such a circumstance but I don't think we want to push that burden to clients to handle and we certainly don't want a failure to log a change to fail out the parent flight.

Any idea what `changes` logs as in the log entry?

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlight.java
```diff
@@ -56,6 +58,7 @@ public CreateProfileFlight(FlightMap inputParameters, Object applicationContext)
       // we can link the profile to the MRG only after the Sam resource has been created
       addStep(new LinkBillingProfileIdToMrgStep(samService, billingProfile, user));
     }
-    addStep(new CreateProfileFinishStep());
+    addStep(new RecordProfileCreationStep(changeLogDao));
+    addStep(new CreateProfileFinishStep(changeLogDao));
```
aherbst-broad, line 62: Both of the above steps look like they're recording a profile creation event, is that intentional?

service/src/main/java/bio/terra/profile/db/ProfileChangeLogDao.java
```diff
@@ -0,0 +1,154 @@
...
+  public Optional<UUID> recordProfileCreate(BillingProfile profile) {
+    var sql =
+        INSERT_INTO_TABLE
+            + " (profile_id, change_type, change_by, change_date) VALUES (:profile_id, :change_type, :change_by, :change_date)";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue(PROFILE_ID, profile.id())
+            .addValue(CHANGE_TYPE, ChangeLogEntry.ChangeTypeEnum.CREATE.name())
+            .addValue(CHANGE_DATE, profile.createdTime().atOffset(ZoneOffset.UTC))
+            .addValue(CHANGE_BY, profile.createdBy());
```
aherbst-broad, line 72: Is createdBy the user's email address? Is it possible to capture their sam user id instead? I'm on a mini-crusade to stop capturing user email for these sorts of logs, as that is not a reliable proxy for user identity in our system (i.e., think linked microsoft and google logins).

service/src/test/java/bio/terra/profile/db/ProfileChangeLogDaoTest.java
```diff
@@ -0,0 +1,83 @@
...
+            CloudPlatform.AZURE,
+            Optional.empty(),
+            Optional.of(UUID.randomUUID()),
+            Optional.of(UUID.randomUUID()),
+            Optional.of("mrg_id"),
+            profileCreateTime,
+            profileCreateTime,
+            user);
+    var recordId = dao.recordProfileCreate(profile);
+    var records = dao.getChangesByProfile(profile.id());
```
aherbst-broad, line 41: Formatting nit, linebreaks for the usual "arrange/act/assert" sections would help readability.

service/src/main/resources/api/service_openapi.yaml
```diff
@@ -598,6 +598,32 @@ components:
         organization:
           $ref: '#/components/schemas/Organization'
 
+    ChangeLogEntry:
```
aherbst-broad, line 601: This doesn't seem to be used in an API controllers, is this here so we can get an auto generated model for the DAO? If so, an internal representation should be used instead which can then be marshaled to API models down the line (something like [this](https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/src/main/java/bio/terra/profile/service/profile/model/BillingProfile.java#L14).

service/src/main/java/bio/terra/profile/db/ProfileChangeLogDao.java
```diff
@@ -0,0 +1,154 @@
...
+  @Autowired
+  public ProfileChangeLogDao(NamedParameterJdbcTemplate jdbcTemplate, ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  @ReadTransaction
+  public List<ChangeLogEntry> getChangesByProfile(UUID profileId) {
+    var params = new MapSqlParameterSource().addValue(PROFILE_ID, profileId);
+    return jdbcTemplate.query(SQL_GET_BY_PROFILE_ID, params, new ChangeLogMapper(objectMapper));
```
aherbst-broad, line 59: This will almost certainly be consumed by user's for rendering into a list (someday). Given that, a default ASC or DESC ordering by CHANGE_DATE would be good.

service/src/main/java/bio/terra/profile/db/ProfileChangeLogDao.java
```diff
@@ -0,0 +1,154 @@
+package bio.terra.profile.db;
+
+import bio.terra.common.db.ReadTransaction;
+import bio.terra.common.db.WriteTransaction;
+import bio.terra.common.exception.SerializationException;
+import bio.terra.profile.model.ChangeLogEntry;
```
aherbst-broad, line 6: Similar to my other comment: This appears to be an API model, an internal representation should be used at the DAO layer and then marshaled to an API representation further up the stack.

aherbst-broad (commented):
Looking good!

service/src/main/java/bio/terra/profile/db/ProfileChangeLogDao.java
blakery: It's just whatever the `Map.toString()` gives.

I agree that failing because of an issue with changelog serialization would be counterproductive.

service/src/main/java/bio/terra/profile/service/profile/flight/create/CreateProfileFlight.java
blakery: Nope - I just forgot to remove the dao from the `CreateProfileFinishStep` constructor, when I moved the rest of the logic to `RecordProfileCreationStep`.

service/src/main/java/bio/terra/profile/db/ProfileChangeLogDao.java
blakery: Making that change was actually what caused me to realize the issue with service accounts acting as themselves, and needing to specify the original user.

service/src/main/resources/api/service_openapi.yaml
blakery: I added an internal model. 
I left this here, since we intend on adding the endpoints imminently.

aherbst-broad (approved):

settings.gradle
```diff
@@ -1,4 +1,4 @@
 rootProject.name = 'terra-billing-profile-manager'
 include('service', 'client')
 
-gradle.ext.releaseVersion = '0.1.566-SNAPSHOT'
+gradle.ext.releaseVersion = '0.1.567-SNAPSHOT'
```
marctalbott, line 4: Doesn't this bump automatically?

blakery: I think that was from rebasing master - I didn't set it manually.

marctalbott: Huh, I'm just confused about why the diff is showing any change then as merging a PR to main shouldn't change this

service/src/main/java/bio/terra/profile/service/iam/SamService.java
```diff
@@ -110,6 +107,28 @@ public boolean isAuthorized(
...
+  public void verifyResourceAdmin(AuthenticatedUserRequest user, SamAction action)
+      throws InterruptedException {
+    String accessToken = user.getToken();
+    var adminApi = samAdminApi(accessToken);
+    try {
+      var isAuthorized =
+          SamRetry.retry(
+              () ->
+                  adminApi.resourceTypeAdminPermission(
+                      SamResourceType.RESOURCE_ADMIN.getSamResourceName(),
```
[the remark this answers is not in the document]
blakery: You are correct. There was a conversation I had with Doug that made me think otherwise, but I managed to actually try it out in dev while impersonating the rawls service user.

marctalbott (approved):

PR #507, opened 2024-12-12
CORE-69: upgrade sonarqube plugin and github actions

davidangb:
this PR supersedes #504

* upgrade gradle version (via `./gradlew wrapper --gradle-version=8.11.1`, no manual changes). I ran into errors with the gradle cache with the new sonarqube; apparently this is a bug in gradle which was fixed by upgrading gradle.
* upgrade various github actions. I originally thought the sonar scan problem was an issue with github caches. I don't think it is, but I left these upgrades in place because they seem good.
* replace explicit usages of `actions/cache` to cache gradle libs with the newer `setup-gradle` action, which has caching built in. See previous bullet point for explanation.
* upgrade sonarqube plugin version. the whole point of this PR.
* add constraints for `commons-compress` and `commons-io` for the buildSrc classpath to satisfy sonarqube needs. This was the fix to get sonar scans working. The trick was realizing that 1) the plugin was using the `buildScriptClasspath` for its dependencies, and 2) that BPM has a `buildSrc/build.gradle` file dedicated to configuring that classpath. Here, the web-based gradle build scans really helped; I could search for `commons-compress` and see which configuration - turned out to be `buildScriptClasspath` - had an old version.

the other thing that helped was getting a sonar token from Vault so I could successfully run sonar scans locally while tweaking dependencies.


samanehsan (approved):
Yay! Thanks for fixing this! 🎉

dvoet (approved):

PR #535, opened 2025-03-21
CORE-376: turn off double-tracing and 'failed to export spans' error

davidangb:
Update TCL to get its fix for the "failed to export spans" error.

See https://github.com/DataBiosphere/terra-common-lib/pull/229

Tested by running locally, seeing no errors logged, and seeing traces reported to GCP.

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -181,11 +181,6 @@ otel:
   sdk:
     disabled: false # set to true to disable all open telemetry features
 
-  # terra-common-lib configures trace exporting to GCP; we need to disable the
-  # built-in otel exporter.
-  traces:
-    exporter: none
-
   resource:
     attributes:
       service:
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -181,9 +181,13 @@ otel:
   sdk:
     disabled: false # set to true to disable all open telemetry features
 
-  springboot:
-    resource:
-      attributes:
-        service:
-          name: ${spring.application.name}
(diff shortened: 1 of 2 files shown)


service/src/main/resources/application.yml
```diff
@@ -181,9 +181,13 @@ otel:
...
-  springboot:
-    resource:
-      attributes:
-        service:
-          name: ${spring.application.name}
-          version: ${bpm.version.gitTag:unknown}
+  # terra-common-lib configures trace exporting to GCP; we need to disable the
+  # built-in otel exporter.
+  traces:
+    exporter: none
```
davidangb, line 187: this is the relevant fix, as copied from TSPS

service/src/main/resources/application.yml
```diff
@@ -181,9 +181,13 @@ otel:
...
+
+  resource:
+    attributes:
+      service:
+        name: ${spring.application.name}
+        version: ${bpm.version.gitTag:unknown}
```
davidangb, line 193: this is unrelated to the exporter fix, but I noticed the attributes under `otel.springboot.resource.attributes` were not working. The correct key is `otel.resource.attributes`, so I moved them.

dvoet (approved):
LGTM, is there any way to include this in TCL?

pshapiro4broad:
Looking at PRs into TCL I see https://github.com/DataBiosphere/terra-common-lib/pull/192 which seems relevant.

kevinmarete (approved):

dvoet (approved):

